PR #3, opened 2022-10-18
Add first Zephyr example

arnopo:
Install environment for zephyr examples by:
1) porting the  [samples/subsys/ipc/openamp_rsc_table sample](https://github.com/OpenAMP/openamp-zephyr-staging/tree/main/samples/subsys/ipc/openamp_rsc_table)
2) updating it to be able to replay the https://github.com/OpenAMP/openamp-system-reference/wiki/Samples-and-demos on the stm32mp157F-dK2 board


examples/zephyr/rpmsg_multi_services/src/main_remote.c
```diff
@@ -0,0 +1,386 @@
...
+static K_SEM_DEFINE(data_tty_sem, 0, 1);
+
+static void platform_ipm_callback(const struct device *dev, void *context,
+				  uint32_t id, volatile void *data)
+{
+	LOG_DBG("%s: msg received from mb %d\n", __func__, id);
+	k_sem_give(&data_sem);
+}
+
+static int rpmsg_recv_cs_callback(struct rpmsg_endpoint *ept, void *data,
```
edmooring, line 107: Why is it rpmsg_recv_cs_callback() and data_sc_sem? Is there a difference in meaning between "cs" and "sc"?

examples/zephyr/rpmsg_multi_services/src/main_remote.c
```diff
@@ -0,0 +1,386 @@
+/*
+ * Copyright (c) 2020, STMICROELECTRONICS
```
TanmayShah-xilinx, line 2: All the files contains license year 2020. Does it need 2022 ?
Rest of it looks good to me.

west.yml
```diff
@@ -13,5 +13,5 @@ manifest:
 
   projects:
     - name: openamp-zephyr-staging
-      revision: v3.2.0
+      revision: aefc06df32e419ca5e122727b57e2051d462cf00
```
edmooring, line 16: This doesn't correspond to the commit message. Is this something peculiar to Zephyr?

edmooring (commented):
Not being a Zephyr expert, this is a less than substantive review.

edmooring (approved):
Looks good to go.

TanmayShah-xilinx (approved):
Looks good to me.

wmamills:
I will provide suggested changes later today.
There are two main ones in the west manifest.
* use main as the revision to avoid bumps every day
* (We can have a different manifest for release tags)
* rename openamp-zephyr-staging to zephyr on clone to keep the west usage simple

wmamills:
I double checked and the only thing I modified was the manifest.  What I have looks like this at the end:
  projects:
    - name: openamp-zephyr-staging
      path: zephyr
      revision: main
      import: true

With these changes you can change the readme to eliminate the -z openamp-zephyr-staging from the west command line.

arnopo:
> I double checked and the only thing I modified was the manifest. What I have looks like this at the end:
> 
> ```
>   projects:
>     - name: openamp-zephyr-staging
>       path: zephyr
>       revision: main
>       import: true
> ```
> 
> With these changes you can change the readme to eliminate the -z openamp-zephyr-staging from the west command line.

arnopo:
> I will provide suggested changes later today. There are two main ones in the west manifest.
> 
> * use main as the revision to avoid bumps every day
This means that we would have to set the `main `branch of the on top of the Zephyr revision we want to use, to be aligned with a release. Or do you propose that we use the main branch of the zephyr main repo?

> * (We can have a different manifest for release tags)

> * rename openamp-zephyr-staging to zephyr on clone to keep the west usage simple

Right that would simplify the use, I will test it

PR #12, opened 2023-07-12
Refactor example code

glneo:
This helps make these example files shorter and less intimidating for new users. Useful when trying to showcase OpenAMP to new users and teams.

### DIFF
--- a/examples/linux/common/common.c
+++ b/examples/linux/common/common.c
@@ -84,6 +84,7 @@ int bind_rpmsg_chrdev(const char *rpmsg_dev_name)
 	if (ret < 0) {
 		fprintf(stderr, "Failed to write %s to %s, %s\n",
 			rpmsg_chdrv, fpath, strerror(errno));
+		close(fd);
 		return -EINVAL;
 	}
 	close(fd);
@@ -101,6 +102,7 @@ int bind_rpmsg_chrdev(const char *rpmsg_dev_name)
 	if (ret < 0) {
 		fprintf(stderr, "Failed to write %s to %s, %s\n",
 			rpmsg_dev_name, fpath, strerror(errno));
+		close(fd);
 		return -EINVAL;
 	}
 	close(fd);
--- a/examples/linux/common/common.c
+++ b/examples/linux/common/common.c
@@ -7,6 +7,7 @@
 #include <linux/rpmsg.h>
 #include <stddef.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
 #include <sys/ioctl.h>
 #include <unistd.h>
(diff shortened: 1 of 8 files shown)


examples/linux/rpmsg-echo-test/echo_test.c
```diff
@@ -24,6 +24,8 @@
 #include <string.h>
 #include <linux/rpmsg.h>
 
+#include "../common/common.h"
```
arnopo, line 27: you can remove the include of<sys/ioctl.h>

arnopo:
@TanmayShah-xilinx : could you confirm that the PR does not introduce regression on your platform?

examples/linux/common/common.c
```diff
@@ -0,0 +1,187 @@
...
+	if (fd < 0) {
+		fprintf(stderr, "Failed to open %s, %s\n",
+			fpath, strerror(errno));
+		return -EINVAL;
+	}
+	ret = write(fd, rpmsg_chdrv, strlen(rpmsg_chdrv) + 1);
+	if (ret < 0) {
+		fprintf(stderr, "Failed to write %s to %s, %s\n",
+			rpmsg_chdrv, fpath, strerror(errno));
+		return -EINVAL;
```
arnopo, line 87: Could you take opportunity to better manage error case?
in case of error in this function,  would be better to close the file before return

examples/linux/common/common.c
```diff
@@ -0,0 +1,187 @@
...
+	while ((ent = readdir(dir)) != NULL) {
+		if (!strncmp(ent->d_name, "rpmsg_ctrl", 10)) {
+			sprintf(dpath, "/dev/%s", ent->d_name);
+			closedir(dir);
+			printf("open %s\n", dpath);
+			fd = open(dpath, O_RDWR | O_NONBLOCK);
+			if (fd < 0) {
+				fprintf(stderr, "open %s, %s\n",
+					dpath, strerror(errno));
+				return fd;
```
arnopo, line 133: same in this function the directory is not closed

arnopo (changes requested):

examples/linux/rpmsg-echo-test/Makefile
```diff
@@ -1,6 +1,6 @@
 
 APP = echo_test
-APP_OBJS = echo_test.o
+APP_OBJS = echo_test.o ../common/common.o
```
TanmayShah-xilinx, line 3: Thanks for your patch.

We should include common.o in clean target as well.
So when we do make clean, it re-builds common.o as well.
or probably introduce new target (distclean), that will remove common.o as well. clean will only remove local directory *.o and distclean will remove all dependencies as well such as common.o. I am fine with either approach.

Thanks.

TanmayShah-xilinx (commented):
LGTM

TanmayShah-xilinx:
> @TanmayShah-xilinx : could you confirm that the PR does not introduce regression on your platform?

@arnopo I verified matrix multiply and proxy app demo. It is working fine without introducing regressions on zcu102 qemu platform. I have some comments, but apart from that it looks good to me.

examples/linux/common/common.c
glneo: Looks like it is closed on line 127 above.

examples/linux/common/common.c
glneo: Right now this is an exact copy/paste without change from before. I'll add a new patch on top to fix this.

examples/linux/rpmsg-echo-test/Makefile
glneo: I'll add 'distclean'. I'm not sure how the 'clean' option would interact when one wanted only to clean the current project.

Fixing out of tree builds (or using CMake or similar) would fix all that, but that would be a task for another day :)

TanmayShah-xilinx (approved):

examples/linux/common/common.c
arnopo: I had missed it, thanks for pointing it out.

arnopo (approved):
LGTM

PR #33, opened 2023-11-02
Xilinx DTS: Use upstream source and only change what we need to.

wmamills:
Instead of checking in decompiled dtbs, use the the upstream kernel as the source of DTS.  Then use include files to change what we need to.

* Use the Linux v6.5.8 DTS source files
* Using include files add the openamp specific definitions
* Update example/dts/,gitignore to ignore dtbo files everywhere
* Update xilinx/.gitignore to ignore the specific products of the Makefile
* This handles the current versions of zcu102 and kv260
* This handles lockstep and split mode of the R5 subsystem

The OpenAMP devicetree changes here are the work of Tanmay Shah and the Makefile and restructure to include files is by Bill Mills.


tnmysh (approved):
LGTM.

wmamills:
@arnopo any comment?  This is Xilinx specific content so I assume you are not to concerned.

PR #35, opened 2023-11-17
Several fixes for Zynqmp DTB

wmamills:
- Switch to all overlays
- Add a compatible DTB for Xilinx bare-metal demos
- Fix the mailbox config for split mode (not used yet)


tnmysh (approved):
LGTM.

PR #36, opened 2023-11-18
DTB fixes for v2023.10 release branch

wmamills:
These DTB fixes came after the v2023.10.0 tag and we will not move that tag.
However this fixes are needed fro the v2023.10 demo images so we include them here.
The Zynqmp changes are already on main and are cherry picked back to this branch.
The ST audio disable workaround is only for this branch as hopefully it will not be needed in the future.

* Move zynqmp to all overlays (away from dtsi files)
* Add DTBO and DTBs for R5 bare-metal demos from Xilinx 2023.2
* Fix the mailbox config for zynqmp split mode (not used yet)
* Disable all audio devices in ST to avoid kernel oops in v6.5.8
* Fix the move to overlays for ZCU102

### DIFF
--- a/examples/linux/dts/st/.gitignore
+++ b/examples/linux/dts/st/.gitignore
@@ -1,0 +1,5 @@
+# These files come from a copy command in the Makefile
+# don't check them in even though they look like source files
+
+stm32mp157c-dk2.dts
+.linux-src-check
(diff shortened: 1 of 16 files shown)


wmamills:
@tnmysh this is the same 3 commits we merged to main.  Should be an easy review.

@arnopo this is what I did to get my existing v6.5.8 kernel to boot and run the demo w/o any kernel oops.

As it is the weekend I will be going forward assuming that these commits are OK as is for the release branch.  I do not plan to put the ST workaround on the main branch.

tnmysh (approved):
LGTM.

arnopo (approved):
Ok to have the temporary DT overlays for this release

PR #37, opened 2023-11-20
dtb: Fix zcu102 bare-metal

wmamills:
This fix is already in the v2023.10 branch but it is needed here on main as well.

A holdover from the dtsi model was getting both mailbox configs in the bare-metal dtb for zcu102.

Fix this by making the zcu102-openamp overlay only disable the serial port and not include the zynqmp-openamp overlay defs.


tnmysh (approved):
lgtm

PR #38, opened 2023-12-08
DO-NOT-MERGE: examples: zephyr: add virtio device samples

uLipe:
EDIT: 

This is the solution I reached to move the device side implementation out of zephyr tree and serve it as reference for other OS to implement their own device side, since it depends directly of the target OS/platform.

The idea is to create a zephyr-virtio module out of tree and pushes it to the zephyr specifics examples and create the the device structure required from zephyr from there.

I need to clean-up a bit this PR, for example add the instructions to build at the current stage, please notice the only required step here is to point to the this openamp branch in order to have the virtio_mmio_dev infrastructure available for zephyr usage: 

https://github.com/uLipe/open-amp/tree/feature/virtio_mmio_dev


uLipe:
@arnopo , @danmilea , @wmamills , I opened this PR to show you how I pulled out the zephyr stuff to implement a virtio mmio device side without touching the zephyr tree, but only with open-amp plus zephyr out of tree facilities

uLipe:
@danmilea , I hope this PR helps you to see the full picture of the virtio mmo dev being used.

PR #45, opened 2024-06-28
Unbind rpmsg dev

tnmysh:
Fix linux apps and unbind rpmsg device with driver after rpmsg communication is done.

### DIFF
--- a/examples/linux/common/common.c
+++ b/examples/linux/common/common.c
@@ -66,7 +66,8 @@ char *get_rpmsg_ept_dev_name(const char *rpmsg_char_name,
 int bind_rpmsg_chrdev(const char *rpmsg_dev_name)
 {
 	char fpath[256];
-	char *rpmsg_chdrv = "rpmsg_chrdev";
+	const char *rpmsg_chdrv = "rpmsg_chrdev";
+	char drv_override[64] = {0};
 	int fd;
 	int ret;
 
@@ -74,12 +75,39 @@ int bind_rpmsg_chrdev(const char *rpmsg_dev_name)
 	sprintf(fpath, "%s/devices/%s/driver_override",
 		RPMSG_BUS_SYS, rpmsg_dev_name);
 	printf("open %s\n", fpath);
-	fd = open(fpath, O_WRONLY);
+	fd = open(fpath, O_RDWR);
 	if (fd < 0) {
 		fprintf(stderr, "Failed to open %s, %s\n",
 			fpath, strerror(errno));
 		return -EINVAL;
 	}
+
+	ret = read(fd, drv_override, sizeof(drv_override));
+	if (ret < 0) {
+		fprintf(stderr, "Failed to read %s (%s)\n",
+			fpath, strerror(errno));
+		close(fd);
+		return ret;
+	}
+
+	printf("current drv override = %s\n", drv_override);
+
+	/*
+	 * Check driver override. If "rpmsg_chrdev" string is
+	 * found, then don't attempt to bind. If null string is found,
+	 * then no driver is bound, and attempt to bind rpmsg char driver.
+	 * Any other case, fail binding driver, as device is busy.
+	 */
+	if (strncmp(drv_override, rpmsg_chdrv, strlen(rpmsg_chdrv)) == 0) {
+		close(fd);
+		return 0;
+	} else if (strncmp(drv_override, "(null)", strlen("(null)")) != 0) {
+		fprintf(stderr, "error: device %s is busy, drv bind=%s\n",
+		       rpmsg_dev_name, drv_override);
+		close(fd);
+		return -EBUSY;
+	}
+
 	ret = write(fd, rpmsg_chdrv, strlen(rpmsg_chdrv) + 1);
 	if (ret < 0) {
 		fprintf(stderr, "Failed to write %s to %s, %s\n",


arnopo:
hi @tnmysh:  do you tried to use RPMSG_DESTROY_EPT_IOCTL instead?

tnmysh:
> hi @tnmysh: do you tried to use RPMSG_DESTROY_EPT_IOCTL instead?

Actually, I can do that in separate PR, where binding and unbinding both are taken care via IOCTLs, instead of writing to sysfs nodes. For now this PR is only fixing existing workflow.

arnopo:
> > hi @tnmysh: do you tried to use RPMSG_DESTROY_EPT_IOCTL instead?
> 
> Actually, I can do that in separate PR, where binding and unbinding both are taken care via IOCTLs, instead of writing to sysfs nodes. For now this PR is only fixing existing workflow.

you are right,  the bind is implemented in current exemples, so the unbind make sense

examples/linux/common/common.c
```diff
@@ -109,6 +109,33 @@ int bind_rpmsg_chrdev(const char *rpmsg_dev_name)
...
+	printf("write %s to %s\n", rpmsg_dev_name, fpath);
+	ret = write(fd, rpmsg_dev_name, strlen(rpmsg_dev_name) + 1);
+	if (ret < 0) {
+		fprintf(stderr, "Failed to write %s to %s, %s\n",
+			rpmsg_dev_name, fpath, strerror(errno));
+		close(fd);
+		return -EINVAL;
+	}
+	close(fd);
+	return 0;
```
arnopo, line 136: suggestion
```c
	ret = write(fd, rpmsg_dev_name, strlen(rpmsg_dev_name) + 1);
	close(fd);
	if (ret < 0) {
		fprintf(stderr, "Failed to write %s to %s, %s\n",
			rpmsg_dev_name, fpath, strerror(errno));
		return -EINVAL;
	}

	return 0;
```

arnopo (approved):
LGTM

wmamills:
As discussed in the rpmsg call today, the demos should work if the correct driver is already bound and in this case at least the driver should not be unbound.

If the driver DOES need to be bound, it is debatable if it should be unbound on exit or just leave it knowing that the person that needs something else can do the unbind themselves.

tnmysh:
> As discussed in the rpmsg call today, the demos should work if the correct driver is already bound and in this case at least the driver should not be unbound.
> 

Done.

examples/linux/common/common.c
```diff
@@ -63,6 +63,35 @@ char *get_rpmsg_ept_dev_name(const char *rpmsg_char_name,
...
+	fd = open(fpath, O_WRONLY);
+	if (fd < 0) {
+		fprintf(stderr, "Failed to open %s, %s\n",
+			fpath, strerror(errno));
+		return -EINVAL;
+	}
+	ret = read(fd, rpmsg_chdrv, sizeof(rpmsg_chdrv));
+	if (ret < 0) {
+		fprintf(stderr, "Failed to write %s to %s, %s\n",
+			rpmsg_chdrv, fpath, strerror(errno));
```
arnopo, line 86: it is a read , and in case of error `rpmsg_chdrv `is invalid

`fprintf(stderr, "Failed to read %s (%s)\n", fpath, strerror(errno));`

examples/linux/common/common.c
```diff
@@ -63,6 +63,35 @@ char *get_rpmsg_ept_dev_name(const char *rpmsg_char_name,
...
+		close(fd);
```
arnopo, line 87: the `close(fd)` can be moved just after `ret = read(fd, rpmsg_chdrv, sizeof(rpmsg_chdrv));` to avoid to duplicate it

examples/linux/rpmsg-echo-test/echo_test.c
```diff
@@ -120,9 +120,12 @@ int main(int argc, char *argv[])
 		fprintf(stderr, "access(%s): %s\n", fpath, strerror(errno));
 		return -EINVAL;
 	}
-	ret = bind_rpmsg_chrdev(rpmsg_dev);
-	if (ret < 0)
-		return ret;
+
+	if (!is_rpmsg_chrdev_bind(rpmsg_dev)) {
```
arnopo, line 124: The function is_rpmsg_chrdev_bind can return true, false or an error. Seems that the error case should be treated

arnopo (changes requested):

examples/linux/common/common.c
tnmysh: Ack.

examples/linux/common/common.c
tnmysh: Ack.

examples/linux/rpmsg-echo-test/echo_test.c
tnmysh: Ack. Thanks for catching this.

tnmysh: Also, I think is_rpmsg_chrdev_bind should be moved within bind_rpmsg_chrdev.
As it's common to all the apps.

tnmysh: Done.

examples/linux/common/common.c
```diff
@@ -66,20 +66,48 @@ char *get_rpmsg_ept_dev_name(const char *rpmsg_char_name,
...
+	 * Check driver override. If "rpmsg_chrdev" string is
+	 * found, then don't attempt to bind. If null string is found,
+	 * then no driver is bound, and attempt to bind rpmsg char driver.
+	 * Any other case, fail binding driver, as device is busy.
+	 */
+	if (strncmp(drv_override, rpmsg_chdrv, strlen(rpmsg_chdrv)) == 0) {
+		close(fd);
+		return 0;
+	} else if (strncmp(drv_override, "(null)", strlen("(null)")) != 0) {
+		printf("error: device %s is busy, drv bind=%s\n",
```
arnopo, line 105: fprintf?

arnopo (approved):
with a minor comment

examples/linux/common/common.c
tnmysh: done.

tnmysh:
@edmooring , @wmamills requesting reviews.

tnmysh:
@wmamills , created follow up issue here: https://github.com/OpenAMP/openamp-system-reference/issues/50

wmamills:
I have tested this with openamp-demo.
I first versified the failure mode:
docker run -it --rm openamp/demo-lite
qemu-zcu102 demo3
(login as root)
./demo3A
echo_test

This 2nd echo test fails. Likewise for demo3B and demo3C.

I then built a new image with openamp-ci-builds including this PR and updated the demo.
With the new image echo_test and mat_mul_demo can run multiple times after their associated demo script.

However ./demo3C followed by proxy_app continues to fail.
I belive this is because the proxy app firmware destroys its endpoint afterward and is not releated to this PR but this needs confirmation.

tnmysh:
@wmamills yes I confirm. There will be separate PR to fix that behavior.

Thanks.

wmamills (approved):
I have tested using openamp-demo demo3

PR #43, opened 2024-05-20
update for V2024.05 release

arnopo:
Update the zephyr version and point to the v2024.05 release candidate

PR #46, opened 2024-07-02
west: update to zephyr 3.6

arnopo:
Update to the last Zephyr release

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -16,7 +16,7 @@ manifest:
   projects:
     - name: zephyr
       remote: zephyr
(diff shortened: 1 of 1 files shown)


wmamills:
@arnopo Sorry I did not see this one.  I cherry picked the first commit of PR #43 and pushed that.  If you look at that one you will see it is only 1 commit now not two.

I suggest we close this PR.  Otherwise I can force push to take the other one off and merge this one.  I don't think we need a big review as PR #43 was reviewed.

PR #47, opened 2024-07-02
Zynqmp DTS: Update for v6.6 kernel

wmamills:
No real update was needed for v6.6 but update the names and comments to make things more clear.  Delete the #if 0 stuff in the bare-metal dtso.

Looking ahead this configuration looks compatible and still needed for v6.6 through at least v6.10 (as of v6.10-rc6).


wmamills:
@arnopo FYI on this one as it is all Xilinx.  I am using upstream DTBs for ST this release as the audio issue is gone for v6.6.
(The ST DTB stuff was never in main; it was only in the v2023.10 branch)

wmamills:
Actually runtime tests of this are failing.  I can run rpmsg_multi_services (Zephyr 3.5 & 3.6) and kernel 6.6 with the old dtb and it works OK.  When I switch to this new dtb I get no rpmsg activity.

I will be looking at this more closely.

wmamills:
@tnmysh disregard the comment above.  The code does work as is and this PR is good to go.
Please review as soon as you can.

tnmysh (approved):
LGTM.

arnopo (approved):

PR #60, opened 2024-10-25
legacy_apps: Fix rpc demo dynamic library build

arnopo:
Add the missing -fPIC option for dynamic library generation to fix the following error:

/usr/bin/ld: CMakeFiles/rpc_demod-shared.dir/__/__/system/linux/machine/generic/rsc_table.c.o: relocation R_X86_64_PC32 against symbol `resources' can not be used when making a shared object; recompile with -fPIC


tnmysh (approved):
LGTM.

iuliana-prodan (approved):

PR #65, opened 2025-02-15
docs: update the readthedocs integration for PRs in this repo

wmamills:
Upstream Read The Docs has tightened up the schema checking and this broke what we were doing for PRs.

A good number of things were tried but the best compromise is to:
* supply a dummy sphinx config file in this repo
* override the default build command to use the openamp-docs dir

Also:
* add more documentation about what is going on
* lock to html only format.  PRs only do html anyway but make sure

### DIFF
--- a/.readthedocs-conf.py
+++ b/.readthedocs-conf.py
@@ -1,0 +1,14 @@
+# Although RTDs now requires sphinx: configuration: it does not pass it to the
+# build command and it still looks for conf.py in the current dir.  It is
+# confusing to have a conf.py that will only work in the very specific RTDs
+# dir structure so we don't do that.
+
+# So we supply a no-op config file and override the build command to run in the
+# openamp-docs dir
+
+# Since we have our own build command now, we could eliminate this file
+# completely but that also disables the predefined sphinx steps of
+# create_environment and install.  The current setup keeps the build
+# environment closer to that used by the real openamp-docs build
+
+print(".readthedocs-conf.py does nothing")
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,5 +1,16 @@
+# The main readthedocs build is in openamp-docs and includes this repo as a
+# sub-module.
+
+# This config file supports doc preview for PRs in this repo.
+# This allows us to see how a PR in this repo will look in the docs before it
+# is promoted to main and before the openamp-docs repo picks up the new version.
+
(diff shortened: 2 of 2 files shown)


wmamills:
@arnopo @tnmysh I am assuming that the docs that are now in open-amp/doc/apps will be moved here to docs/apps.
If this is not true it is a bit silly to create the doc dir only to hold a do nothing file.
Let me know what you think

tnmysh:
Should we open PR to track that migration?

wmamills:
@tnmysh yes, please open an issue
@arnopo do you both agree that this is the right thing to do?  Move them as is from open-amp to openamp-system-reference?
The other alternative is to move the directly to openamp-docs.

tnmysh:
> @tnmysh yes, please open an issue in open-amp @arnopo do you both agree that this is the right thing to do? Move them as is from open-amp to openamp-system-reference? The other alternative is to move the directly to openamp-docs.

I think keeping apps related docs in system-reference will be more helpful. Then openamp-docs can point link to this repo. But I am open to otherway as well.

arnopo:
> @tnmysh yes, please open an issue in open-amp @arnopo do you both agree that this is the right thing to do? Move them as is from open-amp to openamp-system-reference? The other alternative is to move the directly to openamp-docs.

I would be in favor of having complex documentation (for instance, PNG schematics) in the openamp-docs repo and just a README in the system-reference demo explaining how to build and run the demo.

And regarding @sipke's work on documentation, we probably have some duplications that should be cleaned up in the openamp-docs repo.

tnmysh:
open-amp issue is here: https://github.com/OpenAMP/open-amp/issues/638

tnmysh (approved):

wmamills:
Based on the discussion at OpenAMP system reference meeting on Feb 19th, we don't expect the doc/ directory in this repo to conatin anything.  I don't think we should create it just for this one file.

I suggest I rework this PR to move the doc/readthedocs_conf.py file to /.readthedocs_conf.py (in the root dir and with a leading dot)

@arnopo when you get back from vacation let me know if you agree

wmamills:
@arnopo I have moved the config doc to the root dir (with a leading . ) and I don't create the doc dir.
I have tested this in my own account and it works OK.
I have removed the RFC status on this.

arnopo (approved):

PR #67, opened 2025-04-02
legacy apps: update cmake minimum support to 3.16

arnopo:
We currently support an old version of CMake (3.0.2) that was released in 2018. Next cmake revisions will support 3.5 as minimum version.

Update the minimum version to 3.16, aligned with the libmetal and the open-amp libraies.

By updating to CMake 3.16 as the minimum version, we can also clean up the following policy add-ons:
- CMP0048 added in CMake v3.0
- CMP0077 added in CMake v3.13

### DIFF
--- a/examples/legacy_apps/CMakeLists.txt
+++ b/examples/legacy_apps/CMakeLists.txt
@@ -1,11 +1,4 @@
-cmake_minimum_required (VERSION 3.0.2)
-if (POLICY CMP0048)
-  cmake_policy(SET CMP0048 NEW)
-endif()
-
-if (POLICY CMP0077)
-  cmake_policy(SET CMP0077 NEW)
-endif()
+cmake_minimum_required (VERSION 3.16)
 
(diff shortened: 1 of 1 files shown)


arnopo:
@tnmysh @edmooring : Please could you have look?
Without this patch the OpenampCI fails:

 -- Build legacy Apps --
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.

wmamills (approved):
LGTM

edmooring (approved):
Good to go.

PR #71, opened 2025-04-23
Fix assertion and free

arnopo:
- Fix assertion introduced by https://github.com/OpenAMP/open-amp/pull/644
- fix memory free reported by asan


edmooring (commented):
Looks good to go.

PR #74, opened 2025-05-05
examples: dts: xlnx: update dtso for remoteproc nodes

tnmysh:
Use new remoteproc nodes introduced in 6.12 kernel.

### DIFF
--- a/examples/linux/dts/xilinx/zynqmp-openamp.dtso
+++ b/examples/linux/dts/xilinx/zynqmp-openamp.dtso
@@ -58,7 +58,7 @@
 		};
 	};
 
-	zynqmp_ipi {
+	zynqmp-ipi {
 		#address-cells = <2>;
 		#size-cells = <2>;
 		ranges;
@@ -67,7 +67,7 @@
 			reg = <0x00 0xff990040 0x00 0x20>,
 			      <0x00 0xff990060 0x00 0x20>,
 			      <0x00 0xff990200 0x00 0x20>,
-			      < 0x00 0xff990220 0x00 0x20>;
+			      <0x00 0xff990220 0x00 0x20>;
 			reg-names = "local_request_region",
 				    "local_response_region",
 				    "remote_request_region",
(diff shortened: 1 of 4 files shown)


examples/linux/dts/xilinx/zynqmp-split.dtso
```diff
@@ -3,7 +3,10 @@
...
 
-&remoteproc {
-	xlnx,cluster-mode = <0>;
+&rproc_lockstep {
+	status = "disabled";
 };
 
+&rproc_split {
+	status = "okay";
+};
```
arnopo, line 12: Regarding other dtso , it not clear to me what is the default status for `rproc_lockstep` and `rproc_split` nodes. I would suggest to to either apply this in other dtso or  just overwrite status when necessary.

tnmysh: So default in kernel repo both nodes are disabled, and in each board it's enabled as needed.

tnmysh: Actually I checked. For zcu102 board The split mode is enabled by default. For all other boards both the nodes are disabled by default. So, I think, I will keep split mode enabled for all the boards and apply lockstep via dtso only when needed.

tnmysh: I will push new changes that will make it clear.

arnopo (approved):

examples/linux/dts/xilinx/zynqmp-openamp.dtso
```diff
@@ -90,7 +90,25 @@
...
+		r5f@1 {
+			memory-region = <&rproc_1_fw_image>, <&rpu1vdev0vring0>,
+					<&rpu1vdev0vring1>, <&rpu1vdev0buffer>;
+			mboxes = <&ipi_mailbox_rpu1 0>, <&ipi_mailbox_rpu1 1>;
+			mbox-names = "tx", "rx";
+		};
+	};
+
+	rproc_split: remoteproc-split {
+		status = "okay";
```
iuliana-prodan, line 111: This can be skipped. If the `status` property is missing from a node, it defaults to `"okay"`.

tnmysh: Thanks for your reviews. For some boards the split node is disabled in kernel repository dtsi. That's why it needs to be enabled via this overlay here. SO status = okay is needed.

iuliana-prodan (approved):

PR #76, opened 2025-05-16
Amd may16 2025 support freertos

bentheredonethat:
@arnopo  here is new attempt at bringing in freertos support for legacy apps
CC @tnmysh


examples/legacy_apps/examples/echo/freertos/main.c
```diff
@@ -0,0 +1,84 @@
...
+#include <openamp/version.h>
+#include <metal/alloc.h>
+#include <metal/log.h>
+#include <metal/version.h>
+#include "platform_info.h"
+#include "rpmsg-echo.h"
+#include "FreeRTOS.h"
+#include "task.h"
+
+#define LPRINTF(format, ...) xil_printf(format, ##__VA_ARGS__)
```
tnmysh, line 15: use metal_err. Please take care of this for each PR you have to open from here on. We don't really need any other way to print logs. Only metal_* macros.

examples/legacy_apps/examples/rpc_demo/CMakeLists.txt
```diff
@@ -26,6 +26,13 @@ endif (${PROJECT_SYSTEM} STREQUAL "linux")
 foreach (_app ${app_list})
   collector_list (_sources APP_COMMON_SOURCES)
   list (APPEND _sources "${CMAKE_CURRENT_SOURCE_DIR}/${_app}.c")
+  if (${_app} STREQUAL rpc_demo)
+    if (${PROJECT_SYSTEM} STREQUAL "linux")
+        list (APPEND _sources "${CMAKE_CURRENT_SOURCE_DIR}/generic/main.c")
```
tnmysh, line 31: Arnaud had pointed this before. What is importance of generic/main.c under Linux OS? Can you please check which demo is using this option?

examples/legacy_apps/examples/matrix_multiply/CMakeLists.txt
```diff
@@ -16,6 +16,13 @@ set (OPENAMP_LIB open_amp)
 foreach (_app matrix_multiply matrix_multiplyd)
   collector_list (_sources APP_COMMON_SOURCES)
   list (APPEND _sources "${CMAKE_CURRENT_SOURCE_DIR}/${_app}.c")
+  if (${_app} STREQUAL matrix_multiplyd)
+    if (${PROJECT_SYSTEM} STREQUAL "linux")
+      list (APPEND _sources "${CMAKE_CURRENT_SOURCE_DIR}/generic/main.c")
```
tnmysh, line 21: Please check which linux demo is using this option and if it's needed, then please add comment explaining it.

bentheredonethat: the OpenAMP Remote will be using this (rpmsg-echo,  matrix_multiplyd and rpc_demo)
Linux build was tested to work.

examples/legacy_apps/examples/rpc_demo/CMakeLists.txt
bentheredonethat: as noted below, Linux remote will use this. confirmed Linux build worked for this.

bentheredonethat: added comment in latest PR for this.

examples/legacy_apps/examples/echo/freertos/main.c
bentheredonethat: will fix.

examples/legacy_apps/examples/echo/generic/main.c
```diff
@@ -0,0 +1,65 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+#include <stdio.h>
+#include <openamp/open_amp.h>
+#include <openamp/version.h>
+#include <metal/alloc.h>
+#include <metal/log.h>
+#include <metal/version.h>
+#include "platform_info.h"
+#include "rpmsg-echo.h"
```
arnopo, line 11: Please reorder using alphabetical order.

examples/legacy_apps/examples/echo/generic/main.c
```diff
@@ -0,0 +1,65 @@
+/*
+ * SPDX-License-Identifier: BSD-3-Clause
```
arnopo, line 2: copyright is missing

examples/legacy_apps/examples/echo/generic/main.c
```diff
@@ -0,0 +1,65 @@
...
+#include <metal/log.h>
+#include <metal/version.h>
+#include "platform_info.h"
+#include "rpmsg-echo.h"
+
+#define LPRINTF(fmt, ...) metal_info("%s():%u " fmt, __func__, __LINE__, ##__VA_ARGS__)
+#define LPERROR(fmt, ...) metal_err("ERROR: " fmt, ##__VA_ARGS__)
+
+/* Application entry point */
+int main(int argc, char *argv[])
```
arnopo, line 17: I would expect that this function (same for other examples) is just a cut/past of the one in `rpmsg-echo.c`

examples/legacy_apps/examples/echo/CMakeLists.txt
```diff
@@ -21,6 +21,13 @@ foreach (_app rpmsg-echo-ping rpmsg-echo)
     list (APPEND _sources "${CMAKE_CURRENT_SOURCE_DIR}/rpmsg-ping.c")
   elseif (${_app} STREQUAL "rpmsg-echo")
     list (APPEND _sources "${CMAKE_CURRENT_SOURCE_DIR}/rpmsg-echo.c")
+    if (${PROJECT_SYSTEM} STREQUAL "linux")
+      # Note that only OpenAMP Remote (rpmsg-echo) will use this branch.
+      # The generic/main routine is for both baremetal and linux targets.
+      list (APPEND _sources "${CMAKE_CURRENT_SOURCE_DIR}/generic/main.c")
+    else()
+      list (APPEND _sources "${CMAKE_CURRENT_SOURCE_DIR}/${PROJECT_SYSTEM}/main.c")
+    endif(${PROJECT_SYSTEM} STREQUAL "linux")
```
arnopo, line 30: @wmamills : does this generate some regression in your CI?

examples/legacy_apps/examples/rpc_demo/CMakeLists.txt
arnopo: the term  generic is used for the baremetal in libmetal. and this approach could generate some regression for non Linux system.

Regarding this commit I wonder if we should not just keep `main()` in the example using `weak`  attribute and here append `${PROJECT_SYSTEM}/main.c `if the file exists. The  ${PROJECT_SYSTEM}/main.c` would redefine the `main()`

examples/legacy_apps/examples/echo/generic/main.c
bentheredonethat: will fix.

examples/legacy_apps/examples/echo/generic/main.c
bentheredonethat: will fix.

examples/legacy_apps/examples/echo/rpmsg-echo.c
```diff
@@ -46,6 +50,11 @@ static int rpmsg_endpoint_cb(struct rpmsg_endpoint *ept, void *data, size_t len,
 	if (rpmsg_send(ept, data, len) < 0) {
 		LPERROR("rpmsg_send failed\r\n");
 	}
+
+#ifdef FREERTOS_BSP
```
arnopo, line 54: is it a standard config or an AMD config?

examples/legacy_apps/examples/echo/rpmsg-echo.c
```diff
@@ -46,6 +50,11 @@ static int rpmsg_endpoint_cb(struct rpmsg_endpoint *ept, void *data, size_t len,
...
+	xTaskResumeFromISR(rpmsg_task);
```
arnopo, line 55: I did not see that the task was suspended. Do I missed something?
This function should be generic if needed. with an implementation per system

examples/legacy_apps/system/freertos/machine/zynqmp_r5/gic_init.c
```diff
@@ -0,0 +1,32 @@
+/*
```
arnopo, line 1: Please Split this commit, one to introduce free RTOS, one for AMD implementation

bentheredonethat:
Hi @arnopo  i have updated the PR as follows:

1. split generic rename demo logic and GIC init
2. split freertos demos and freertos GIC init
3. add stubs that make app-suspend for freertos suspend/resume generic 

style fixes

examples/legacy_apps/system/freertos/machine/zynqmp_r5/gic_init.c
```diff
@@ -0,0 +1,32 @@
...
+#include "platform_info.h"
+#include <stdio.h>
+#include <stdarg.h>
+#include "xil_exception.h"
+#include "xil_printf.h"
+#include "xparameters.h"
+#include "xscugic.h"
+
+/* Interrupt Controller setup */
+int app_gic_initialize(void)
```
arnopo, line 22: Would it be possible to move this in example/legacy_apps/ machine? 
Would be nice that the system folder does not contain machine specificities.
This request is mainly for the FreeRTOS support that is new .

examples/legacy_apps/system/freertos/machine/zynqmp_r5/app_task.c
```diff
@@ -0,0 +1,24 @@
...
+#include "FreeRTOS.h"
+#include "platform_info.h"
+#include "task.h"
+
+#ifdef FREERTOS_BSP
+extern TaskHandle_t rpmsg_task;
+
+void app_suspend(void)
+{
+vTaskSuspend(rpmsg_task);
```
arnopo, line 17: tabulation

examples/legacy_apps/system/freertos/machine/zynqmp_r5/app_task.c
```diff
@@ -0,0 +1,24 @@
...
+}
+
+void app_resume(void)
+{
+	xTaskResumeFromISR(rpmsg_task);
+}
+#endif /* FREERTOS_BSP */
```
arnopo, line 24: seems to me that is file could be generic changing function name and passing task as 
 a void * argument

arnopo (commented):
Much better, thanks for the split.
The last point I would like to improve is to avoid to have machine specificity ( if possible) in the `system/freertos`
I would expect that routine in this folder is just a wrapper to Freertos API

bentheredonethat:
Hi @arnopo  i updated so that hte gic_init logic is (a) only there for r5 and (b) i have used format similar to earlier suggestion where the gic_init and app-resume logic specific to each OS is only there in the zynqmp_r5 area.

bentheredonethat:
@arnopo  i updated the app-task logic to be generic/same for freertos and baremetal... but im not sure its better than it was before because we still have to do an extern to expose the rpmsg-task wherever the suspend/resume APIs are called.

arnopo:
> > rpmsg-task
> 
> What about adding rmpsg_task_suspend/resume(void) function in main in main.c that will call vTaskSuspend/Resume? and creating a main.h that would be include in the machine if neded to suspend/resume the FreeRTOS rpmsg_task?

@bentheredonethat 
I don't see your last comment received by mail, so I add it here too answer

>>What about adding rmpsg_task_suspend/resume(void) function in main in main.c that will call vTaskSuspend/Resume? and creating a main.h that would be include in the machine if neded to suspend/resume the FreeRTOS rpmsg_task?


> Hi @arnopo
>i'm missing something here - could you explain this?

Yes sorry it is not cristal clear. The main principle is that the  suspend/resume of the `rpmsg_task` task seems to be generic.
Regarding your implementation you need to call it in zynqmp_r5_a53_rproc.c while you call `_rproc_wait` in zynqmp_r5/platform_info.c

As you mention the previous implementation I proposed implies dependency between the sample and the platform to export  `rpmsg_task`
 
>
> the catch here is that the platform_poll() and zynqmp_r5_a53_proc_irq_handler() routines need to have the rpmsg_task_suspend/resume calls.
> If I understand your suggestion is it to:
> (a) define main.h with weak implementations

add follwing function in main.c
```
void rpmsg_app_suspend(void)
{
	vTaskSuspend(rpmsg_task);
}

void rpmsg_app_resume(void)
{
	xTaskResumeFromISR(rpmsg_task);
}
```
Create a main.h that declare these functions

> (b) keep commit legacy_apps: zynqmp_r5: Add default usage of app_suspend/resume but it will now also include main.h

- come back to the call of `_rproc_wait` in `zynqmp_r5/platform_info.c`
- include the main.h and call r`pmsg_app_suspend`/`rpmsg_app_resume` in `zynqmp_r5_a53_rproc.c `

> (c) define new freertos-amd specific implementations in the freertos main.c files for each demo
> just want to make sure i understand the suggestion here
> 

I hope this explanation clarifies things. Don't hesitate to reach out if anything is still unclear or if you see an issue with my proposal.

bentheredonethat:
Hi @arnopo  apprecate you the time to explain here.

The one concern I have is that we have 3 different main routines per OS. So I would be adding that same freertos implementation to the 3 legacy apps' freertos main.c files and likewise for generic main.c files.

That is why I added the implementaion in separate file, so that the implementation is not duplicated.

If you still want me to proceed in doing the implementation in each of the legacy app' s main.c files I will do so with a 'weak' attribute and test that.

thank you!

arnopo:
> Hi @arnopo apprecate the time to explain here.

Thank you for taking the time to consider my suggestions. 😃

> 
> The one concern I have is that we have 3 different main routines per OS. So I would be adding that same freertos implementation to the 3 legacy apps' freertos main.c files and likewise for generic main.c files.

Your point is valid !

- For the generic main is seems not needed as you call `_rproc_wait` in your routines, so you could come back to your initial implementation removing app_suspend/resume, right?
- For the FreeRTos: yes this this the draw back. from my point of view i would prefer to duplicate than hide OS dependent code in the machine.

What about creating a examples/legacy_apps/system/freertos/thread.c file to communalise the routine that could be in future reused for some other OSes? 
- Add  `thread_TaskCreate function (task_ID, task_func_ , task_name, stack_size, priority, p_created_task)`
    => the `task_ID` would be an index of  a table of static TaskHandle_t tasks, and defined as an enum in thread.h
```
    TaskHandle_t  task_kist[THREAD_APP_NB_TASK_ID];
```
- Add `thread_task_suspend` and `thread_task_resume` functions with task ID provided as argument.
- create a thread.h 
   - containing thread routine declarations
   - containing an enum that list task IDs 
```
enum {
    THREAD_APP_TASK_ID = 0,
    THREAD_APP_NB_TASK_ID,
} thread_task_id;
``` 

> 
> That is why I added the implementaion in separate file, so that the implementation is not duplicated.
> 
> If you still want me to proceed in doing the implementation in each of the legacy app' s main.c files I will do so with a 'weak' attribute and test that.
> 
> thank you!

edmooring (approved):
The amount of duplicated code in this PR bothers me, but I don't see an easy way to refactor it at the moment. Since these are legacy apps, and therefore unlikely to be modified, I think we can let it go.

examples/legacy_apps/examples/rpc_demo/generic/main.c
```diff
@@ -25,6 +26,18 @@
 #define LPRINTF(format, ...) metal_info(format, ##__VA_ARGS__)
 #define LPERROR(format, ...) metal_err("ERROR: " format, ##__VA_ARGS__)
 
+void rpmsg_app_suspend(void *data)
+{
+	(void)data;
+	_rproc_wait();
```
arnopo, line 32: `_rproc_wait` is AMD specific ,  it should not be call here, 

I would make this function empty or just remove `rpmsg_app_suspend`and  `rpmsg_app_resume` for generic implementation if unused

arnopo (changes requested):

bentheredonethat:
Hi @arnopo  you mentioned 
"_rproc_wait is AMD specific , it should not be call here,

I would make this function empty or just remove rpmsg_app_suspendand rpmsg_app_resume for generic implementation if unused"

and 

"If rpmsg_app_suspend and rpmsg_app_resume are removed for generic, this commit become useless"


thanks for calling these out. we use ARM instruction wfi here in _rproc_wait.

it is small but _rproc_wait is used in this case because of that instruction.

Also I might be missing something here -- but the _rproc_wait is in fact replaced in a later commit by rpmsg_app_suspend so it does 'go away'. Unless this coment is calling something else out here?

I agree with the sentimenet that these routines are small and in the generic context it makes sense to keep them as is.

BUT the real utility of those comes in only for the freertos implementation. The generic implementation is there for consistency and convention.

Given that context, please advise on what to change here, I am unsure of the action t for this suspend/resume part on this PR.

arnopo:
> Hi @arnopo you mentioned "_rproc_wait is AMD specific , it should not be call here,
> 
> I would make this function empty or just remove rpmsg_app_suspendand rpmsg_app_resume for generic implementation if unused"
> 
> and
> 
> "If rpmsg_app_suspend and rpmsg_app_resume are removed for generic, this commit become useless"
> 
> thanks for calling these out. we use ARM instruction wfi here in _rproc_wait.
> 
> it is small but _rproc_wait is used in this case because of that instruction.
> 
> Also I might be missing something here -- but the _rproc_wait is in fact replaced in a later commit by rpmsg_app_suspend so it does 'go away'. Unless this coment is calling something else out here?
> 
> I agree with the sentimenet that these routines are small and in the generic context it makes sense to keep them as is.
> 
> BUT the real utility of those comes in only for the freertos implementation. The generic implementation is there for consistency and convention.
> 
> Given that context, please advise on what to change here, I am unsure of the action t for this suspend/resume part on this PR.

My bad [examples/legacy_apps/machine/zynqmp_r5/platform_info.c](https://github.com/OpenAMP/openamp-system-reference/pull/76/files#diff-1f7c261e2b925054f236b7efc0d6ed75209fbb4054c955145c13135b327643f0) is comon for freeRtos and generic

I implemented a solution that should work. At least it works on my Linux machine ( demo on PC)
Could you test it and tell me if it is ok for you?
My branch based on you PR: https://github.com/arnopo/openamp-system-reference/tree/app_suspend
if ok for you fill free to update your PR with that ( fixing potential bugs)

bentheredonethat:
Hi @arnopo  thanks for the collaboration here! LGTM on those 2 patches. see updated branch below.

examples/legacy_apps/machine/zynqmp_r5/platform_info.c
```diff
@@ -54,8 +55,7 @@
 #define SHARED_BUF_OFFSET 0x8000UL
 #endif /* !SHARED_BUF_OFFSET */
 
-#ifndef RPMSG_NO_IPI
-#define _rproc_wait() asm volatile("wfi")
+
```
arnopo, line 58: Remove added blanck line

examples/legacy_apps/machine/zynqmp/platform_info.c
```diff
@@ -108,9 +108,6 @@ static struct remoteproc rproc_inst;
 extern int init_system(void);
```
arnopo, line 108: If you squash the patches in one please also squash the commit removing 
```
Signed-off-by: Arnaud Pouliquen <arnaud.pouliquen@foss.st.com>

examples: legacy: Add system suspend/resume
```
in the middle of the commit message

arnopo (approved):
2 minor comments, else LGTM

bentheredonethat:
@arnopo  udpated!

arnopo:
@edmooring : could you have a look please

edmooring (approved):
Looks good to go.

PR #78, opened 2025-06-30
Amd june30 recovery update

bentheredonethat:
This PR adds support in the AMD RPU Legacy OpenAMP so that the RPU can have IPC reconnected if host has an uncontrolled reboot.
CC @tnmysh

### DIFF
--- a/examples/legacy_apps/machine/zynqmp_r5/rsc_table.c
+++ b/examples/legacy_apps/machine/zynqmp_r5/rsc_table.c
@@ -17,7 +17,11 @@
 /* Place resource table in special ELF section */
 #define __section_t(S)          __attribute__((__section__(#S)))
 #define __resource              __section_t(.resource_table)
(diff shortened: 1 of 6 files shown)


examples/legacy_apps/machine/zynqmp_r5/rsc_table.c
```diff
@@ -64,6 +66,24 @@ void *get_resource_table (int rsc_id, int *len)
 {
 	(void) rsc_id;
 	*len = sizeof(resources);
+	/* make copy of resources to restore later */
+	if (!initial_resources) {
+		initial_resources = (struct remote_resource_table *)malloc(*len);
+		memset(initial_resources, 0, *len);
+		memcpy(initial_resources, &resources, *len);
```
arnopo, line 85: The memset seems useless as overwritten by the mempy()

tnmysh (approved):

PR #84, opened 2025-09-27
zynqmp_r5: use libmetal extension module

tnmysh:
Xlnx BSP was moved out of libmetal to new library module maintained in downstream repo: libmetal_xlnx_extension.a

Modify zynqmp_r5 platform to use interrupt controller defined in this new library module.

In the process fix freertos compilation on zynqmp_r5 platform as well.


examples/legacy_apps/system/freertos/CMakeLists.txt
```diff
@@ -1,7 +1,4 @@
 # Required libraries for Demos for linking.
-find_library(FIND_FREERTOS freertos)
-if (FIND_FREERTOS)
-  collect(PROJECT_LIB_DEPS freertos)
-endif(FIND_FREERTOS)
+collect(PROJECT_LIB_DEPS "libfreertos.a")
```
arnopo, line 2: seems strange that you have to change the Freertos integration method. the original declaration seems to me more generic

tnmysh: @arnopo 
Actually you are right.` collect(PROJECT_LIB_DEPS <text>)` ends up as linker flags `-l<text>` I should keep it as before. Other places as well. I will fix it.

examples/legacy_apps/machine/zynqmp_r5/platform_info.c
```diff
@@ -84,7 +84,11 @@ struct metal_device kick_device = {
 	.node = {NULL},
 #ifndef RPMSG_NO_IPI
 	.irq_num = 1,
+#ifdef FREERTOS_BSP
+	.irq_info = (void *)(IPI_IRQ_VECT_ID - 32),
```
arnopo, line 88: nitpicking :  The value `32` seems a magic number here. I could be better to use a define or a config linked to FreeRTOS.

arnopo (approved):
just a minor remark else LGTM

arnopo (approved):

PR #92, opened 2025-12-29
zephyr: add versalnet_rpu support

bentheredonethat:
The Versal-NET RPU board uses MBOX so in this demo use IPM over MBOX driver for that board.


arnopo (approved):
LGTM

edmooring (approved):
Looks good to go.

PR #94, opened 2026-01-29
examples: legacy_apps: xlnx: Remove xlnx_hw_to_bsp_irq()

bentheredonethat:
The Lopper-generated amd_platform_info.h already accounts for BSP. So remove the unneeded function xlnx_hw_to_bsp_irq() which tries to account for IRQ system number difference from baremetal (generic) and FreeRTOS BSPs.

### DIFF
--- a/examples/legacy_apps/machine/xlnx/README.md
+++ b/examples/legacy_apps/machine/xlnx/README.md
@@ -1,23 +1,36 @@
 # Steps to generate inputs for AMD-Xilinx RPU Firmware Demos
 
-Dependencies:
-1. Lopper : https://github.com/devicetree-org/lopper.git
-2. System Device Tree generated from design : https://docs.amd.com/r/en-US/ug1647-porting-embeddedsw-components/Generating-a-System-Device-Tree-Using-SDTGen
+Below is sample run for SOM KV260 platform
 
-Below is sample run for Versal Gen 1 platform
-## Generate OpenAMP RPU Device Tree
+## Pick up System Device Tree
 
+```sh
+wget https://edf.amd.com/sswreleases/rel-v2025.2/sdt/2025.2/2025.2_1115_1_11150857/external/k26-smk-kv-sdt/k26-smk-kv-sdt_2025.2_1115_1_11150857.tar.gz
+tar xvf k26-smk-kv-sdt_2025.2_1115_1_11150857.tar.gz
+export SDT=$PWD/k26-smk-kv-sdt_2025.2_1115_1_11150857/system-top.dts
+```
+
+## Set up Lopper
+
+```sh
+git clone https://github.com/devicetree-org/lopper.git
+python3 -m venv .venv
+source .venv/bin/activate
+pip3 install -r lopper/requirements.txt
+pip3 install lopper
+
+export LOPPER_PY=$PWD/lopper/lopper.py
+```
+
+## Apply Domain YAML to System Device Tree
 SDT is the System Device Tree generated from design
+
 ```sh
 export LOPPER_DTC_FLAGS="-b 0 -@"
+python3 $LOPPER_PY -f --enhanced   -x '*.yaml' -i $YAML $SDT yaml_applied.dts
+python3 $LOPPER_PY -f --enhanced yaml_applied.dts rpu.dts -- gen_domain_dts psu_cortexr5_0   --openamp_no_header
 
-python3 lopper.py -f --enhanced \
-  -x '*.yaml' \
-  -i $YAML $SDT yaml_applied.dts
-
-python3 lopper.py -f --enhanced \
-  yaml_applied.dts rpu.dts \
-  -- gen_domain_dts psu_cortexr5_0   --openamp_no_header
+export RPU_DTS=$PWD/rpu.dts
 ```
 The above Device Tree "rpu.dts" will be used for configuration of the app's interrupts, shared memory and linker script.
 
@@ -25,14 +38,8 @@ The above Device Tree "rpu.dts" will be used for configuration of the app's inte
 
 ```sh
 export LOPPER_DTC_FLAGS="-b 0 -@"
-export CONFIG_DTFILE=rpu.dts
-
-cd openamp-system-reference/examples/legacy_apps/machine/zynqmp_r5
-python3 lopper.py -O -f -v --enhanced  --permissive \
-  -O . ${CONFIG_DTFILE} -- openamp --openamp_header_only \
-  --openamp_output_filename=amd_platform_info.h \
-  --openamp_remote=psv_cortexr5_0
-cd -
+python3 lopper.py -O -f --enhanced  --permissive  -O . ${RPU_DTS} -- openamp --openamp_header_only \
+ --openamp_output_filename=amd_platform_info.h --openamp_remote=psu_cortexr5_0
 ```
 The output amd_platform_info.h needs to be in the location denoted above of "openamp-system-reference/examples/legacy_apps/machine/zynqmp_r5" BEFORE
 cmake configure step.
@@ -41,8 +48,6 @@ cmake configure step.
 
 ```sh
 export LOPPER_DTC_FLAGS="-b 0 -@"
-export CONFIG_DTFILE=rpu.dts
-python3 lopper.py -O ${S} rpu.dts \
-  -- baremetallinker_xlnx psv_cortexr5_0 <output location> openamp
+python3 lopper.py -O . $RPU_DTS -- baremetallinker_xlnx psu_cortexr5_0 . openamp
 ```
 The RPU Application Linker config object needs to be pointed to with cmake variable LINKER_METADATA_FILE at cmake configure step.
--- a/examples/legacy_apps/machine/xlnx/README.md
+++ b/examples/legacy_apps/machine/xlnx/README.md
@@ -1,23 +1,48 @@
 # Steps to generate inputs for AMD-Xilinx RPU Firmware Demos
 
-Dependencies:
-1. Lopper : https://github.com/devicetree-org/lopper.git
-2. System Device Tree generated from design : https://docs.amd.com/r/en-US/ug1647-porting-embeddedsw-components/Generating-a-System-Device-Tree-Using-SDTGen
+Below is sample run for SOM KV260 platform
 
-Below is sample run for Versal Gen 1 platform
-## Generate OpenAMP RPU Device Tree
+## Pick up Domain YAMLs
 
+```sh
+git clone https://github.com/Xilinx/meta-xilinx.git -b rel-v2026.1
+export OPENAMP_OVERLAY_YAML=$PWD/meta-xilinx/meta-xilinx-standalone-sdt/conf/domainyaml/zynqmp-openamp-overlay.yaml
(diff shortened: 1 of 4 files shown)


tnmysh:
@bentheredonethat lopper is not widely used yet. So we should handle both cases i.e. with and without lopper.

Can we implement a way to know if generated header file is used or hard-code definition from the header file ?
I remember we had _AMD_GENERATED_ macro passed previously. Can we use it?

bentheredonethat:
Hi @tnmysh 

yes its the AMD_GENERATED header file symbol, no?

In other words:
https://github.com/OpenAMP/openamp-system-reference/blob/main/examples/legacy_apps/machine/xlnx/CMakeLists.txt#L40

If we want some extra awareness without Lopper, why not push that to BSP logic? e.g. metal_xlnx_extension

tnmysh:
> Hi @tnmysh
> 
> yes its the AMD_GENERATED header file symbol, no?
> 
> In other words: https://github.com/OpenAMP/openamp-system-reference/blob/main/examples/legacy_apps/machine/xlnx/CMakeLists.txt#L40
> 
> If we want some extra awareness without Lopper, why not push that to BSP logic? e.g. metal_xlnx_extension


@bentheredonethat good point. Yes I had tried it. But that doesn't work because the zynqmp driver ops is also using the irq_info.
Here: https://github.com/OpenAMP/openamp-system-reference/blob/4bdabbb5b6aaaa902f56d408c091dbefc11e62ca/examples/legacy_apps/machine/xlnx/zynqmp/zynqmp_linux_r5_proc.c#L138

If we move irq conversion logic to BSP, then irq_info above is not passing correct irq information, and interrupt handler is not called.

bentheredonethat:
Hi @tnmysh  thanks for the comment -- I might be missing something but I dont think 'zynqmp' area applies. 

- The routine xlnx_hw_to_bsp_irq () is not called in the 'zynqmp' area. 
- The 'zynqmp' section is for Linux. 
- The Lopper generated header is only used for baremetal/generic and FreeRTOS.

So it seems to me like this commit does not conflict at all but again I might be missing something here.

tnmysh:
> Hi @tnmysh thanks for the comment -- I might be missing something but I dont think 'zynqmp' area applies.
> 
>     * The routine xlnx_hw_to_bsp_irq () is not called in the 'zynqmp' area.
> 
>     * The 'zynqmp' section is for Linux.
> 

This file is for r5: https://github.com/OpenAMP/openamp-system-reference/blob/4bdabbb5b6aaaa902f56d408c091dbefc11e62ca/examples/legacy_apps/machine/xlnx/zynqmp/zynqmp_linux_r5_proc.c

It maps irq_info interrupt number to isr routine in the *_init function.

>     * The Lopper generated header is only used for baremetal/generic and FreeRTOS.
> 
> 
> So it seems to me like this commit does not conflict at all but again I might be missing something here.

bentheredonethat:
@tnmysh  Im still confused. Historically the 'zynqmp' subdir is for linux. you can see that a few ways:

1. it has a reference DTSI
2. i nthe file you reference it says 'Linux r5 remoteproc' where Linux is first https://github.com/OpenAMP/openamp-system-reference/blob/main/examples/legacy_apps/machine/xlnx/zynqmp/zynqmp_linux_r5_proc.c#L142
3. it is using printf (not xil_printf)
4. also we have used 'zynqmp' as convention for linux side in libmetal, openamp demos historically - i can provide pointers there.

also as further reasoning (said this in earlier comment)  - there is no mention of the routine in the init function for this subdir https://github.com/OpenAMP/openamp-system-reference/blob/main/examples/legacy_apps/machine/xlnx/zynqmp/zynqmp_linux_r5_proc.c#L66 

can you please clarify how the routine xlnx_hw_to_bsp_irq() is being called in this subdir or why this commit causes an issue? 

I understand that xlnx_hw_to_bsp_irq() was setting up the IRQ INFO based on BSP. But for the zynqmp/linux subdir - this is not called.

The routine xlnx_hw_to_bsp_irq() is presently called only from xlnx_machine_init(). The routine xlnx_machine_init() is only called in https://github.com/OpenAMP/openamp-system-reference/blob/main/examples/legacy_apps/machine/xlnx/zynqmp_r5/platform_info.c. So from my looking through this - there is no link to the 'zynqmp' subdir that affects the removal of the xlnx_hw_to_bsp_irq() call.

Also the Lopper-flow is fully upstream and the README at https://github.com/OpenAMP/openamp-system-reference/tree/main/examples/legacy_apps/machine/xlnx shows how to generate the Lopper header file.

tnmysh:
Hello @bentheredonethat ,

Thanks for pointing above mistake. I had posted link to a wrong file. Here is the actual file that is used for R5: https://github.com/OpenAMP/openamp-system-reference/blob/4bdabbb5b6aaaa902f56d408c091dbefc11e62ca/examples/legacy_apps/machine/xlnx/zynqmp_r5/zynqmp_r5_a53_rproc.c#L88

My concern was we can't move irq conversion from HW number to BSP number to metal extension because above place is using _irq_info_. And the conversion must happen before it.

We discussed offline, that we will be move to system-device-tree flow by default. And hence, old way of hardcoding of interrupt numbers are obsolete, and in that case this PR is valid.

I don't mind merging this PR once we have following:

1) If Vivado output XSA is accessible to everyone via public domain.

2) People have proper documentation on how to generate required header file from that XSA.
  2.a) XSA -> System-device-tree generation using sdt-gen. 
  2.b) system-device-tree to  amd_platform_info.h header file generation via lopper.

Once we have above two, I will merge this PR in the upstream repo.

Thank You.

bentheredonethat:
Hi @tnmysh  thanks for your comment 

The README for the xlnx vendor shows this already  -- https://github.com/OpenAMP/openamp-system-reference/tree/main/examples/legacy_apps/machine/xlnx 

Is that sufficient?

bentheredonethat:
@tnmysh  updated with README content

tnmysh:
Hi @bentheredonethat 

This is still not working.

Few things I suggest:

1) please decide one commit or release tag of the lopper repo, and checkout branch with that commit or tag in the documentaiton.

2) $YAML is not assigned. Assign it to zynqmp related yaml.

After doing both above, I see following error when generating the header. Not sure what's missing:

python3 $LOPPER_PY -O -f --enhanced  --permissive  -O . ${RPU_DTS} -- openamp --openamp_header_only  -
-openamp_output_filename=amd_platform_info.h --openamp_remote=psu_cortexr5_0
 -> xlnx_openamp_parse
 --> openamp_nontree_outputs_handler
[INFO]: ------> OPENAMP: XLNX:   platform info:  ['xlnx,zynqmp-smk-k26-rev2', 'xlnx,zynqmp-smk-k26-rev1', 'xlnx,zynqmp-smk-k26-revB', 'xlnx,zynqmp-smk-k26-revA', 'xlnx,zynqmp-smk-k26', 'xlnx,zynqmp', 'ZynqMP S
MK-K26 Rev2/1/B/A']
[WARNING]: the assist returned false, check for errors ...

bentheredonethat:
presently waiting for release of system device tree I can point to before taking next steps

nathalie-ckc:
2026-07-15 System Reference call:
Ben to check on status of system DT release to public & post an update to this PR.

bentheredonethat:
@tnmysh  please look at at latest - i have posted using public release artifacts
CC @nathalie-ckc

examples/legacy_apps/machine/xlnx/README.md
```diff
@@ -1,38 +1,52 @@
...
-python3 lopper.py -f --enhanced \
-  -x '*.yaml' \
-  -i $YAML $SDT yaml_applied.dts
+## Set up Lopper
 
-python3 lopper.py -f --enhanced \
-  yaml_applied.dts rpu.dts \
-  -- gen_domain_dts psu_cortexr5_0   --openamp_no_header
+```sh
+git clone https://github.com/devicetree-org/lopper.git
```
tnmysh: We need fix commit of lopper. So, after clone add this step:  `cd lopper; git checkout -b <commit hash>` .

examples/legacy_apps/machine/xlnx/README.md
```diff
@@ -1,38 +1,52 @@
...
-export CONFIG_DTFILE=rpu.dts
-
-cd openamp-system-reference/examples/legacy_apps/machine/zynqmp_r5
-python3 lopper.py -O -f -v --enhanced  --permissive \
-  -O . ${CONFIG_DTFILE} -- openamp --openamp_header_only \
-  --openamp_output_filename=amd_platform_info.h \
-  --openamp_remote=psv_cortexr5_0
-cd -
+python3 $LOPPER_PY --enhanced  --permissive  -O . ${RPU_DTS} -- openamp --openamp_header_only \
+ --openamp_output_filename=amd_platform_info.h --openamp_remote=psu_cortexr5_0
```
tnmysh, line 54: I tried this step, and amd_platform_info.h was generated. 

Now when I modified the YAML, then the new amd_platform_info.h was generated but, corresponding Linux device-tree was not available. I think we should add instructions to generate the linux device-tree as well?

Otherwise, the demo won't work with old linux device-tree.

examples/legacy_apps/machine/xlnx/README.md
```diff
@@ -41,8 +55,6 @@ cmake configure step.
 
 ```sh
 export LOPPER_DTC_FLAGS="-b 0 -@"
-export CONFIG_DTFILE=rpu.dts
-python3 lopper.py -O ${S} rpu.dts \
-  -- baremetallinker_xlnx psv_cortexr5_0 <output location> openamp
+python3 $LOPPER_PY -O . $RPU_DTS -- baremetallinker_xlnx psu_cortexr5_0 . openamp
 ```
 The RPU Application Linker config object needs to be pointed to with cmake variable LINKER_METADATA_FILE at cmake configure step.
```
tnmysh, line 65: Does that mean `Vitis` must be used to compile the demo?

tnmysh:
@bentheredonethat I am expecting documentation similar to this: https://openamp.readthedocs.io/en/v2025.04.0/lopper/demos/openamp/README.html 

We want to enable users to generate artifacts on both sides, and able to use them on real board.

examples/legacy_apps/machine/xlnx/README.md
bentheredonethat: sure will do

examples/legacy_apps/machine/xlnx/README.md
bentheredonethat: @tnmysh  the linux DT generation is different. Also that is scope creep.

that being said I can add separate commit for that sure.

examples/legacy_apps/machine/xlnx/README.md
bentheredonethat: No - why do you say that ? Yocto SHEL workflow supports the build too.

tnmysh: After execution of above step, this artifact was created `Example.cmake`. 

`LINKER_METADATA_FILE` this variable looks like something xlnx specific build system needs.

bentheredonethat: Indeed it is used in the xlnx specific build system - but .cmake files are used to configure .in files. Native CMake supports this.

tnmysh (approved):
LGTM, works as expected.

arnopo (approved):
LGTM

PR #98, opened 2026-03-13
Xlnx add trace buf and update address space

tnmysh:
- Add trace buffer support
- Update current DDR addresses to match latest system device-tree standards
- fix trivial compiler warning during building matrix multiply demo

### DIFF
--- a/examples/legacy_apps/examples/echo/freertos/main.c
+++ b/examples/legacy_apps/examples/echo/freertos/main.c
@@ -10,7 +10,7 @@
 #include <metal/version.h>
 #include <openamp/open_amp.h>
 #include <openamp/version.h>
-#include "platform_info.h"
+#include "platform_info_common.h"
 #include "rpmsg-echo.h"
 #include <stdio.h>
 #include "task.h"
--- a/examples/legacy_apps/machine/xlnx/zynqmp_r5/CMakeLists.txt
+++ b/examples/legacy_apps/machine/xlnx/zynqmp_r5/CMakeLists.txt
@@ -11,7 +11,7 @@ endif(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/${PROJECT_SYSTEM}/gic_init.c")
 
 # If not provided, add default value for resource table
 if(NOT DEFINED ${RSC_TABLE})
-  set(RSC_TABLE 0x3ed00000 CACHE STRING "")
+  set(RSC_TABLE 0x9800000 CACHE STRING "")
 endif(NOT DEFINED ${RSC_TABLE})
 
 set (_linker_script "${CMAKE_CURRENT_SOURCE_DIR}/linker_remote.ld")
--- a/examples/legacy_apps/machine/xlnx/zynqmp_r5/linker_large_text.ld
+++ b/examples/legacy_apps/machine/xlnx/zynqmp_r5/linker_large_text.ld
@@ -15,12 +15,23 @@ _FIQ_STACK_SIZE = DEFINED(_FIQ_STACK_SIZE) ? _FIQ_STACK_SIZE : 1024;
 _UNDEF_STACK_SIZE = DEFINED(_UNDEF_STACK_SIZE) ? _UNDEF_STACK_SIZE : 1024;
 
 /* Define Memories in the system */
-/* TCM size is set to 2*0x20000 for R5 in lockstep mode */
+
+/*
+ * Though the TCM address space can be contiguous, they should be used in 64KB
+ * chunks, as each 64KB chunk is mapped to a different power-domain.
+ */
 MEMORY
 {
-   psu_ddr_S_AXI_BASEADDR : ORIGIN = 0x3ED00000, LENGTH = 0x00040000
+   /* address space in DDR to store firmware elf sections */
+   psu_ddr_S_AXI_BASEADDR : ORIGIN = 0x9800000, LENGTH = 0x00060000
+
+   /* On-chip SRAM address space to store firmware elf sections */
    psu_ocm_ram_1_S_AXI_BASEADDR : ORIGIN = 0xFFFF0000, LENGTH = 0x10000
+
+   /* ATCM0 address space */
    psu_r5_tcm_ram_0_S_AXI_BASEADDR : ORIGIN = 0x00000, LENGTH = 0x10000
+
+   /* BTCM0 address space */
    psu_r5_tcm_ram_1_S_AXI_BASEADDR : ORIGIN = 0x20000, LENGTH = 0x10000
 }
 
@@ -275,7 +286,7 @@ _SDA2_BASE_ = __sdata2_start + ((__sbss2_end - __sdata2_start) / 2 );
    . += _HEAP_SIZE;
    _heap_end = .;
    HeapLimit = .;
-} > psu_r5_tcm_ram_0_S_AXI_BASEADDR
+} > psu_ddr_S_AXI_BASEADDR
 
 .stack (NOLOAD) : {
    . = ALIGN(16);
--- a/examples/legacy_apps/machine/xlnx/zynqmp_r5/linker_remote.ld
+++ b/examples/legacy_apps/machine/xlnx/zynqmp_r5/linker_remote.ld
@@ -15,12 +15,23 @@ _FIQ_STACK_SIZE = DEFINED(_FIQ_STACK_SIZE) ? _FIQ_STACK_SIZE : 1024;
 _UNDEF_STACK_SIZE = DEFINED(_UNDEF_STACK_SIZE) ? _UNDEF_STACK_SIZE : 1024;
 
 /* Define Memories in the system */
-/* TCM size is set to 2*0x20000 for R5 in lockstep mode */
+
+/*
+ * Though the TCM address space can be contiguous, they should be used in 64KB
+ * chunks, as each 64KB chunk is mapped to a different power-domain.
+ */
 MEMORY
 {
-   psu_ddr_S_AXI_BASEADDR : ORIGIN = 0x3ED00000, LENGTH = 0x00040000
+   /* address space in DDR to store firmware elf sections */
+   psu_ddr_S_AXI_BASEADDR : ORIGIN = 0x9800000, LENGTH = 0x00060000
+
+   /* On-chip SRAM address space to store firmware elf sections */
    psu_ocm_ram_1_S_AXI_BASEADDR : ORIGIN = 0xFFFF0000, LENGTH = 0x10000
+
+   /* ATCM0 address space */
    psu_r5_tcm_ram_0_S_AXI_BASEADDR : ORIGIN = 0x00000, LENGTH = 0x10000
+
+   /* BTCM0 address space */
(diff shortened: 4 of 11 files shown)


examples/legacy_apps/examples/echo/freertos/main.c
```diff
@@ -10,7 +10,7 @@
 #include <metal/version.h>
 #include <openamp/open_amp.h>
 #include <openamp/version.h>
-#include "platform_info.h"
+#include "platform_info_common.h"
```
arnopo, line 13: split in 2 patches one for the  platform agnostic part example on for the AMD/Xilinx  machine update

examples/legacy_apps/machine/xlnx/zynqmp_r5/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ endif(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/${PROJECT_SYSTEM}/gic_init.c")
 
 # If not provided, add default value for resource table
```
arnopo, line 12: commit message need to be reworked

```
legacy_apps: xlnx: update reserved mem regions as per latest carveouts

explain the reason of the update 

Signed-off-by: Tanmay Shah <tanmay.shah@amd.com>
```

examples/legacy_apps/machine/xlnx/zynqmp_r5/platform_info.c
```diff
@@ -248,12 +263,56 @@ platform_create_proc(int proc_index, int rsc_index)
...
+
 static int xlnx_machine_init(void)
 {
 
 	struct metal_init_params metal_param = XLNX_PLATFORM_METAL_INIT_PARAMS;
 	int ret;
 
+	/* Init circular buffer */
+	circ.c_buf = get_rsc_trace_info(&circ.c_len);
+	if (circ.c_buf && circ.c_len)
```
arnopo, line 315:
```suggestion
	if (!circ.c_buf || !circ.c_len) 
		metal_err("invalid trace buffer\n");
```

examples/legacy_apps/machine/xlnx/zynqmp_r5/platform_info.c
```diff
@@ -248,12 +263,56 @@ platform_create_proc(int proc_index, int rsc_index)
...
+	static const char * const level_strs[] = {
+		"emergency",
+		"alert",
+		"critical",
+		"error",
+		"warning",
+		"notice",
+		"info",
+		"debug",
+	};
```
arnopo, line 278: put it before the if condition as it is a variable declaration

examples/legacy_apps/machine/xlnx/zynqmp_r5/platform_info.c
```diff
@@ -248,12 +263,56 @@ platform_create_proc(int proc_index, int rsc_index)
...
+
+	len = sprintf(msg, "%s: ", level_strs[level]);
+
+	if (len < 0 || len > (int32_t)sizeof(msg))
+		len = 0;
```
arnopo: what about just using `snprintf()` and remove test on len?
`len = snprintf(msg, sizeof(msg), "%s: ", level_strs[level]);`

arnopo (changes requested):

examples/legacy_apps/machine/xlnx/zynqmp_r5/platform_info.c
tnmysh: That's good idea. Ack.

examples/legacy_apps/machine/xlnx/zynqmp_r5/platform_info.c
tnmysh: Didn't apply this one, let me know if this is must.

examples/legacy_apps/machine/xlnx/zynqmp_r5/CMakeLists.txt
arnopo, line 9: s/exampples/examples/ in commit message 

That said  you can remove `examples:` as done for other patches

examples/legacy_apps/machine/xlnx/zynqmp_r5/platform_info.c
```diff
@@ -248,12 +263,54 @@ platform_create_proc(int proc_index, int rsc_index)
...
+		"notice",
+		"info",
+		"debug",
+	};
+	char msg[128];
+	char *p;
+	int32_t len;
+	va_list args;
+
+
```
arnopo: useless double blank line

examples/legacy_apps/machine/xlnx/zynqmp_r5/rsc_table.c
```diff
@@ -37,9 +37,11 @@
 #endif /* RING_RX */
 #define VRING_SIZE                  256
 
-#define NUM_TABLE_ENTRIES           1
+#define NUM_TABLE_ENTRIES           2
+#define RSC_TRACE_SZ		    (4*1024)
```
arnopo: nitpicking. The trace contain char, why not just just 
```suggestion
#define RSC_TRACE_SZ		    4096
```

tnmysh:
Hi @arnopo,

I found bug in this patch series when using internal tools, please do not merge it until next update.

Thanks,
Tanmay

tnmysh:
@arnopo, @edmooring  this change is ready for review.

arnopo (approved):

examples/legacy_apps/machine/xlnx/zynqmp_r5/linker_large_text.ld
```diff
@@ -18,7 +18,7 @@ _UNDEF_STACK_SIZE = DEFINED(_UNDEF_STACK_SIZE) ? _UNDEF_STACK_SIZE : 1024;
 /* TCM size is set to 2*0x20000 for R5 in lockstep mode */
 MEMORY
 {
-   psu_ddr_S_AXI_BASEADDR : ORIGIN = 0x3ED00000, LENGTH = 0x00040000
+   psu_ddr_S_AXI_BASEADDR : ORIGIN = 0x9800000, LENGTH = 0x00060000
```
edmooring, line 26: This now conflicts with the comment in line 18 above, Size has changed from 0x40000 to 0x60000 without changing the comment. Which is correct?

examples/legacy_apps/machine/xlnx/zynqmp_r5/linker_remote.ld
```diff
@@ -18,7 +18,7 @@ _UNDEF_STACK_SIZE = DEFINED(_UNDEF_STACK_SIZE) ? _UNDEF_STACK_SIZE : 1024;
 /* TCM size is set to 2*0x20000 for R5 in lockstep mode */
 MEMORY
 {
-   psu_ddr_S_AXI_BASEADDR : ORIGIN = 0x3ED00000, LENGTH = 0x00040000
+   psu_ddr_S_AXI_BASEADDR : ORIGIN = 0x9800000, LENGTH = 0x00060000
```
edmooring, line 26: Same as above. Either the comment in line 18 is wrong or the code in line 21 is wrong.

examples/legacy_apps/machine/xlnx/zynqmp_r5/platform_info.c
```diff
@@ -101,9 +101,9 @@
 #endif /* versal */
 #ifndef SHARED_MEM_PA
 #if XPAR_CPU_ID == 0
-#define SHARED_MEM_PA  0x3ED40000UL
+#define SHARED_MEM_PA  0x9860000UL
```
edmooring, line 104: Again, why the size change from 0x40000 to 0x60000? This also applies to line 106 below.

examples/legacy_apps/machine/xlnx/zynqmp_r5/platform_info.c
```diff
@@ -184,6 +184,21 @@ int xlnx_hw_to_bsp_irq(int sys_irq);
...
+/* circular buffer for trace buffer */
+static struct {
+	char *c_buf;
+	uint32_t c_len;
+	uint32_t c_pos;
+	uint32_t c_cnt;
+} circ;
+
+static void rsc_trace_putchar(char c)
+{
```
edmooring, line 196: A very small nit:
The circular buffer will be left in a logically inconsistent state after c_len calls to rsc_trace_puchar(), where c_pos actually points past the end of the buffer. This might cause problems for readers of the buffer. Switching the order so the put to the buffer is done first, and then the check for the end of the buffer would keep the state of the buffer logically consistent between calls to rsc_trace_putchar().

examples/legacy_apps/machine/xlnx/zynqmp_r5/platform_info.c
```diff
@@ -248,12 +263,54 @@ platform_create_proc(int proc_index, int rsc_index)
...
+		"notice",
+		"info",
+		"debug",
+	};
+	char msg[128];
+	char *p;
+	int32_t len;
+	va_list args;
+
+	if (level > METAL_LOG_DEBUG)
```
edmooring, line 288: This needs an explanatory comment. Why are we escalating the warning level?

edmooring (changes requested):
This comes down to two questions:

Which is the correct size for the TCM lockstep region, 0x40000 or 0x 60000?
Why is the log level escalated for levels above "debug"?

examples/legacy_apps/machine/xlnx/zynqmp_r5/linker_large_text.ld
tnmysh: @edmooring, actually the memory modified here is not TCM, but DDR. So the firmware code can be split and some sections can be loaded into TCM, and some are loaded in the DDR.

The TCM memory is represented here with: psu_r5_tcm_ram_0_S_AXI_BASEADDR

Which I think is 64KB. I agree that the comment should be modified, as the TCM is 64-KB chunks. I will update comment.

examples/legacy_apps/machine/xlnx/zynqmp_r5/linker_remote.ld
tnmysh: Same explanation as above.

examples/legacy_apps/machine/xlnx/zynqmp_r5/platform_info.c
tnmysh: This SHARED_MEM_PA is different region than mentioned in the linker script.

This is where vring buffers are located. This address is moved in the new spec.

examples/legacy_apps/machine/xlnx/zynqmp_r5/platform_info.c
tnmysh: Ack.

tnmysh:
@edmooring addressed all the comments as discussed.

arnopo:
@edmooring, I merged this one because @tnmysh addressed your comment, and I need to release the repository.

If you have other comments, a new PR can be created to address them.

PR #100, opened 2026-03-17
legacy_apps: xlnx: add api to poll apps on vdev status

tnmysh:
In case of the remoteproc repeat attach-detach, RPU firmware should be able to re-create the virtio devices. For this, the firmware should poll on the vdev status, and if the status is 0, then it should destroy the virtio and rpmsg devices. Then it should re-create them and wait for next driver ready status from the linux kernel. The vdev status 0 is usually done during remoteproc detach operation. Mailbox notification is expected from the Linux side on vdev status reset.

### DIFF
--- a/examples/legacy_apps/examples/echo/freertos/main.c
+++ b/examples/legacy_apps/examples/echo/freertos/main.c
@@ -48,18 +48,21 @@ static void rpmsg_listen_task(void *arg)
 	LPRINTF("libmetal lib version: %s\n", metal_ver());
 	LPRINTF("Starting application...\r\n");
 
-	while (1) {
+	do {
 		rpdev = platform_create_rpmsg_vdev(platform, 0,
 						   VIRTIO_DEV_DEVICE,
 						   NULL, NULL);
 		if (!rpdev) {
 			LPERROR("Failed to create rpmsg virtio device.\r\n");
-			break;
+			ret = 0;
 		} else {
-			rpmsg_echo_app(rpdev, platform);
+			ret = rpmsg_echo_app(rpdev, platform);
+			if (ret)
+				metal_err("rpmsg_echo_app failed with err, %d\n",
+					  ret);
 			platform_release_rpmsg_vdev(rpdev, platform);
 		}
-	}
+	} while (!ret);
 
 	LPRINTF("Stopping application...\r\n");
 	platform_cleanup(platform);
--- a/examples/legacy_apps/examples/echo/rpmsg-echo.c
+++ b/examples/legacy_apps/examples/echo/rpmsg-echo.c
@@ -82,16 +82,12 @@ int rpmsg_echo_app(struct rpmsg_device *rdev, void *priv)
 	metal_dbg("RPMsg device TX buffer size: %#x\r\n", rpmsg_get_tx_buffer_size(&lept));
 	metal_dbg("RPMsg device RX buffer size: %#x\r\n", rpmsg_get_rx_buffer_size(&lept));
 
-	while(1) {
-		platform_poll(priv);
-		/* we got a shutdown request, exit */
-		if (shutdown_req) {
-			break;
-		}
-	}
+	ret = platform_poll(priv);
+	if (ret)
+		metal_err("platform poll failed err = %d\n", ret);
 	rpmsg_destroy_ept(&lept);
 
-	return 0;
+	return ret;
 }
 
 /*-----------------------------------------------------------------------------*
--- a/examples/legacy_apps/examples/echo/rpmsg-echo.c
+++ b/examples/legacy_apps/examples/echo/rpmsg-echo.c
@@ -72,7 +72,7 @@ int rpmsg_echo_app(struct rpmsg_device *rdev, void *priv)
 			       rpmsg_service_unbind);
 	if (ret) {
 		LPERROR("Failed to create endpoint.\r\n");
-		return -1;
+		return ret;
 	}
 
 	metal_log(METAL_LOG_NOTICE,
@@ -118,17 +118,21 @@ int __attribute__((weak)) main(int argc, char *argv[])
 
 	LPRINTF("Starting application...\r\n");
 
-	rpdev = platform_create_rpmsg_vdev(platform, 0,
-					   VIRTIO_DEV_DEVICE,
-					   NULL, NULL);
-	if (!rpdev) {
-		LPERROR("Failed to create rpmsg virtio device.\r\n");
-		ret = -1;
-	} else {
-		rpmsg_echo_app(rpdev, platform);
-		platform_release_rpmsg_vdev(rpdev, platform);
+	/* keep repeating application until it fails */
+	do {
 		ret = 0;
-	}
+		rpdev = platform_create_rpmsg_vdev(platform, 0,
+						   VIRTIO_DEV_DEVICE,
+						   NULL, NULL);
+		if (!rpdev) {
+			LPERROR("Failed to create rpmsg virtio device.\r\n");
+			ret = -1;
+		} else {
+			ret = rpmsg_echo_app(rpdev, platform);
+			platform_release_rpmsg_vdev(rpdev, platform);
+		}
+		LPRINTF("Creating vdev devices again\r\n");
+	} while (!ret);
 
 	LPRINTF("Stopping application...\r\n");
 	platform_cleanup(platform);
--- a/examples/legacy_apps/examples/matrix_multiply/generic/main.c
+++ b/examples/legacy_apps/examples/matrix_multiply/generic/main.c
@@ -36,17 +36,19 @@ int main(int argc, char *argv[])
 	LPRINTF("libmetal lib version: %s\n", metal_ver());
 	LPRINTF("Starting application...\r\n");
 
-	rpdev = platform_create_rpmsg_vdev(platform, 0,
-					   VIRTIO_DEV_DEVICE,
-					   NULL, NULL);
-	if (!rpdev) {
-		LPERROR("Failed to create rpmsg virtio device.\r\n");
-		ret = -1;
-	} else {
-		rpmsg_matrix_app(rpdev, platform);
-		platform_release_rpmsg_vdev(rpdev, platform);
-		ret = 0;
-	}
+	/* on vdev reset, recreate rpmsg devices */
+	do {
+		rpdev = platform_create_rpmsg_vdev(platform, 0,
+						   VIRTIO_DEV_DEVICE,
+						   NULL, NULL);
+		if (!rpdev) {
+			LPERROR("Failed to create rpmsg virtio device.\r\n");
+			ret = -EINVAL;
+		} else {
+			ret = rpmsg_matrix_app(rpdev, platform);
+			platform_release_rpmsg_vdev(rpdev, platform);
+		}
+	} while (!ret);
 
 	LPRINTF("Stopping application...\r\n");
 	platform_cleanup(platform);
--- a/examples/legacy_apps/machine/xlnx/zynqmp_r5/platform_info.c
+++ b/examples/legacy_apps/machine/xlnx/zynqmp_r5/platform_info.c
@@ -29,6 +29,7 @@
 #include <metal/sys.h>
 #include <metal/utilities.h>
 #include <openamp/rpmsg_virtio.h>
+#include <openamp/virtio.h>
 
 #include "platform_info.h"
 #include "rsc_table.h"
@@ -414,20 +415,42 @@ err1:
 
 int platform_poll(void *priv)
 {
+	struct rpmsg_virtio_device *rpmsg_vdev;
+	struct remoteproc_virtio *rproc_vdev;
 	struct remoteproc *rproc = priv;
 	struct remoteproc_priv *prproc;
+	struct metal_list *node;
+	uint8_t vdev_status = 0;
 	unsigned int flags;
 	int ret;
 
 	prproc = rproc->priv;
-	while(1) {
+	/* If vdev reset is issued, then stop polling */
+	node = metal_list_first(&rproc->vdevs);
+	if (!node) {
+		metal_err("failed to get rproc_vdev node\n");
+		return -EINVAL;
+	}
+
+	rproc_vdev = metal_container_of(node,
+					struct remoteproc_virtio,
+					node);
+	rpmsg_vdev = rproc_vdev->vdev.priv;
+	if (!rpmsg_vdev) {
+		metal_err("failed to get rpmsg_vdev\n");
+		return -ENODEV;
+	}
+	if (!rpmsg_vdev->vdev) {
+		metal_err("failed to get virtio device\n");
+		return -ENODEV;
+	}
+	do {
 #ifdef RPMSG_NO_IPI
 		if (metal_io_read32(prproc->kick_io, 0)) {
 			ret = remoteproc_get_notification(rproc,
 							  RSC_NOTIFY_ID_ANY);
 			if (ret)
 				return ret;
-			break;
 		}
 		(void)flags;
 #else /* !RPMSG_NO_IPI */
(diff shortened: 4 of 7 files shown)


examples/legacy_apps/examples/echo/rpmsg-echo.c
```diff
@@ -82,13 +82,7 @@ int rpmsg_echo_app(struct rpmsg_device *rdev, void *priv)
...
 	metal_dbg("RPMsg device RX buffer size: %#x\r\n", rpmsg_get_rx_buffer_size(&lept));
 
-	while(1) {
-		platform_poll(priv);
-		/* we got a shutdown request, exit */
-		if (shutdown_req) {
-			break;
-		}
-	}
+	platform_poll_on_vdev_reset(priv, rdev);
```
arnopo: This imposes that platform_poll_on_vdev_reset is defined in all machine.   For instance this breaks build on Linux PC. 

```
/usr/bin/ld: CMakeFiles/rpmsg-echo-shared.dir/rpmsg-echo.c.o: in function `rpmsg_echo_app':
rpmsg-echo.c:(.text+0x37f): undefined reference to `platform_poll_on_vdev_reset'
```
then `platform_poll` seems deprecated

Why not just  keeping platform_poll but adding` struct rpmsg_device *rpdev` parameter?

arnopo (changes requested):

examples/legacy_apps/examples/echo/rpmsg-echo.c
tnmysh: Hmm... But that means, all the current support will move to the virtio reset polling. Is that okay?

Sorry I missed the Linux build on PC.

arnopo: > Hmm... But that means, all the current support will move to the virtio reset polling. Is that okay?
> 

Not sure to understand your point, the implementation of `platform_poll` is already machine dependent.

tnmysh: I mean all for all places where platform_poll is used, is it okay to use platform_poll_on_vdev_reset? Can we deprecate use of platform_poll ? I don't see any reason to keep platform_poll honestly. What do you think?

arnopo: Without `platform_poll`, there is no notification of new incoming rpmsg (see the remoteproc_get_notification() function). It seems to me that your `platform_poll_on_vdev_reset` function is an extension of `platform_poll` with added management of the vdev status reset.

From my point of view, platform_poll is a more generic name. I would update it instead of creating a new function.

In the openamp-system-reference, I don’t think there is a need to apply the deprecation process, as the objective is only to provide reference code.

tnmysh: Okay I will extend `platform_poll`.  Add documentation as well.

tnmysh: @arnopo, I tried to extend platform_poll but it won't work for rpc demo.

Here `rpmsg_rpc_init` API from open-amp library expect platform_poll API with only one argument: https://github.com/OpenAMP/openamp-system-reference/blob/3eec8f826a2f7fd5b3028bfdbe8253e00ec87f0c/examples/legacy_apps/examples/rpc_demo/rpc_demo.c#L64

definition of `rpmsg_rpc_init`: https://github.com/OpenAMP/open-amp/blob/62b28881119277b9d2f244b2314e18bde23c296e/lib/include/openamp/rpmsg_retarget.h#L79

type of API expected: https://github.com/OpenAMP/open-amp/blob/62b28881119277b9d2f244b2314e18bde23c296e/lib/include/openamp/rpmsg_retarget.h#L32

That is why I propose to introduce new API and use it only where it's needed. Let me know if you have any other idea.

tnmysh: Basically, facing this **_warning_**:

 warning: passing argument 7 of 'rpmsg_rpc_init' from incompatib
le pointer type [-Wincompatible-pointer-types]
   64 |                              priv, platform_poll, rpmsg_rpc_shutdown);
      |                                    ^~~~~~~~~~~~~
      |                                    |
      |                                    int (*)(void *, struct rpmsg_device *)

arnopo: From my perspective the issue is that the rpmsg_retarget is more an example that an rpmsg generic service of the open-amp library. The poll should not be done in the service but in the associated openamp-system reference example.

In short term ,  for this particular service, we could update the `rpmsg_rpc_poll` typedef with a note in the release.
In mid, long term,  we should  move the poll from openamp lib to the openamp-system reference

tnmysh: @arnopo Thanks for the suggestion. Even if `rpmsg_retarget` is an example, it is a part of the open-amp library. It is still not part of the system-reference. It is possible, that people have defined their own interfaces like `platform_poll` and passing into `rpmsg_retarget`. IMHO if we decide to change the rpmsg_retarget interface, then we should treat it just same as any other library interface that is exported to the users and go through the deprecation process of an interface. Because as of now, it is still part of the library. 

I think we should aim for long-term solution instead, and remove polling from the service, and do it in the system-referece example instead.

arnopo: > @arnopo Thanks for the suggestion. Even if `rpmsg_retarget` is an example, it is a part of the open-amp library. It is still not part of the system-reference. It is possible, that people have defined their own interfaces like `platform_poll` and passing into `rpmsg_retarget`. IMHO if we decide to change the rpmsg_retarget interface, then we should treat it just same as any other library interface that is exported to the users and go through the deprecation process of an interface. Because as of now, it is still part of the library.
> 
> I think we should aim for long-term solution instead, and remove polling from the service, and do it in the system-referece example instead.

If implementing the long-term solution directly is ok with you, then I am also in favor of this approach.

tnmysh: HI @arnopo,

I did analysis on rpmsg_retarget.c file as well. I think the implementation standpoint it's correct. So, when `open` call is made, it waits for the confirmation from the Host that the file was created. So the polling happens to wait for that confirmation. We can't expect the polling happen in the system-reference app code, because it is required as part of `open` call. Without it the file fd won't be received from the host.

This means, our only option is to change the interface of rpc_poll, and include second argument there.
That also means to go through the deprecation process for that interfaces/typedef.

I think the easiest option instead is, to maintain two polling API in the system reference for now, and deprecate `rpc_poll` later.

tnmysh: HI @arnopo,

I might have a way to achieve the same result, without changing the platform_poll interfaces. I will test it next week and update you. Please ignore above message till then.

Thank you.

tnmysh: @arnopo I pushed the latest changes. Please check. I don't have to change platfrom_poll interface now, and still can poll on vdev flag.

The trick is, I can retrieve vdev from the list of remoteproc vdev devices. Today we have only one remoteproc_vdev for each remoteproc, so I just get the first remoteproc_vdev from the list, and from there I can get virtio device.
Let me know if this looks good.

Thanks.

examples/legacy_apps/examples/echo/freertos/main.c
```diff
@@ -48,18 +48,21 @@ static void rpmsg_listen_task(void *arg)
...
 
-	while (1) {
+	do {
 		rpdev = platform_create_rpmsg_vdev(platform, 0,
 						   VIRTIO_DEV_DEVICE,
 						   NULL, NULL);
 		if (!rpdev) {
 			LPERROR("Failed to create rpmsg virtio device.\r\n");
-			break;
+			ret = 0;
```
arnopo, line 57:
```suggestion
			ret = -1;
```

examples/legacy_apps/examples/matrix_multiply/generic/main.c
```diff
@@ -36,17 +36,19 @@ int main(int argc, char *argv[])
...
-		ret = 0;
-	}
+	/* on vdev reset, recreate rpmsg devices */
+	do {
+		rpdev = platform_create_rpmsg_vdev(platform, 0,
+						   VIRTIO_DEV_DEVICE,
+						   NULL, NULL);
+		if (!rpdev) {
+			LPERROR("Failed to create rpmsg virtio device.\r\n");
+			ret = -EINVAL;
```
arnopo, line 46: sometime  `-1` is returned, sometime  `-EINVAL`.   What about using `EPERM` or `ENXIO` everywhere

examples/legacy_apps/machine/xlnx/zynqmp_r5/platform_info.c
```diff
@@ -438,12 +461,18 @@ int platform_poll(void *priv)
...
+		system_suspend();
 #endif /* RPMSG_NO_IPI */
-	}
+
+		ret = virtio_get_status(rpmsg_vdev->vdev, &vdev_status);
+		if (ret) {
+			metal_err("failed to get vdev status %d\n", ret);
+			return ret;
+		}
+	} while (vdev_status);
```
arnopo, line 474: Except if I miss something this breaks the virtio spec protocol. The virtio driver must not set the status to 0.

```
2.4 Device Reset
The driver may want to initiate a device reset at various times; notably, it is required to do so during device initialization and device cleanup.

The mechanism used by the driver to initiate the reset is transport specific.

2.4.1 Device Requirements: Device Reset
A device MUST reinitialize device status to 0 after receiving a reset.

A device MUST NOT send notifications or interact with the queues after indicating completion of the reset by reinitializing device status to 0, until the driver re-initializes the device.

2.4.2 Driver Requirements: Device Reset
The driver SHOULD consider a driver-initiated reset complete when it reads device status as 0.
```

here the sequence should be

- a reset request should come from the other side
- The machine detect the reset in platform_poll() function and a a specif  error ( perhaps -EPIPE).
- generic application stop the rpmsg and restart it

examples/legacy_apps/examples/echo/rpmsg-echo.c
```diff
@@ -82,16 +82,12 @@ int rpmsg_echo_app(struct rpmsg_device *rdev, void *priv)
...
 	metal_dbg("RPMsg device RX buffer size: %#x\r\n", rpmsg_get_rx_buffer_size(&lept));
 
-	while(1) {
-		platform_poll(priv);
-		/* we got a shutdown request, exit */
-		if (shutdown_req) {
-			break;
-		}
-	}
+	ret = platform_poll(priv);
```
arnopo, line 85: This make break all machines implementation not only the AMD ones.

arnopo (changes requested):

examples/legacy_apps/machine/xlnx/zynqmp_r5/platform_info.c
tnmysh: So as per this virtio standard, this behavior is already broken in the linux kernel side. Because today we see virtio status set to 0 on `detach()` call. Perhaps `rproc_reset_rsc_table_on_detach` this function is restoring the default resource table in the linux kernel which sets vdev_status to 0 on detach.

arnopo: Yes you are right, the Linux should not clean the vdev, status

tnmysh: @arnopo 

How about this. We keep polling up until DRIVER_OKAY bit is set by the linux. If it's not setup anymore, then we stop polling. I think that's not breaking the virtio standard.

tnmysh: > @arnopo
> 
> How about this. We keep polling up until DRIVER_OKAY bit is set by the linux. If it's not setup anymore, then we stop polling. I think that's not breaking the virtio standard.

Actually, that's not a good idea too. Because as per the spec, driver can't choose specific bits to reset.

Instead to reset the device, driver will set the status 0. So, are we not resetting the device when we issue detach() command?

arnopo: From my point of view, we should think with a system-level approach rather than going with a temporary solution.

What about a solution based on a new status in the resource table, as discussed during the last OpenAMP RP call?

We should probably discuss the solution on the Linux kernel side before moving forward with this PR. The fact that Linux does not respect the virtio specification should help us reach a solution.

tnmysh: @arnopo Yes that's what I was thinking too.

The only problem is the backward compatibility from the linux standpoint.

If we implement the new status based on the resource table, and the remote firmware is polling based on current vdev status, then we may break the existing firmwares.

So, I was thinking in the linux side:

if (new_vdev_status in rsc tbl) then:
   do not reset vdev status on detach, instead rely on fw to reset vdev status
else:
  reset vdev status, and notify firmware.
endif

We can discuss this in the next system-reference or rp call. Which comes first.

examples/legacy_apps/examples/echo/rpmsg-echo.c
tnmysh: I don't know how it can break. Remote firmware will still work as it is.

arnopo: Not tested but, , for instance, seems not compatible with following codes:
k
- https://github.com/OpenAMP/openamp-system-reference/blob/main/examples/legacy_apps/system/linux/machine/generic/platform_info.c#L531
- https://github.com/OpenAMP/openamp-system-reference/blob/main/examples/legacy_apps/tests/msg/rpmsg-ping.c#L141

nathalie-ckc:
From 4/15/2026 OpenAMP System Reference call:
* Tanmay will notify Mathieu in the cover letter of the patch that this is violation of virtio spec & continue discussion on mailing list
* It's remote's responsibility when it sees the detach instruction from Linux it should reload clean resource table & set status to 0
* @glneo to review this PR and comment from TI perspective

tnmysh:
Hi @arnopo,

I was exploring the current design. I think the virtio reset notificaiton mechanism via shared memory should be per rproc virtio device. How do you feel about providing reset field in the virtio config space and detect it via one of the virtio feature bits ?

arnopo:
we discussed the point during Linux remoteproc meeting.  As a common solution is require with Linux , let continue discussion in Linux Mailing list using an RFC, to find a common approach

tnmysh:
> we discussed the point during Linux remoteproc meeting. As a common solution is require with Linux , let continue discussion in Linux Mailing list using an RFC, to find a common approach

Thanks @arnopo. Sorry I had posted question before our discussion, and didn't update here. Here is the discussion link: https://lore.kernel.org/all/20260317201251.3920841-1-tanmay.shah@amd.com/

nathalie-ckc:
7/1 OpenAMP System Reference discussion:

Tanmay:
we had discussed this in OpenAMP RP call, and we concluded on two solutions on Linux side.
One, set the status bit to some unique value and notify the remote, and then when remote completes the whole reset cycle, remote will set the status bit to zero.  And that is how Linux knows that remote has completed the reset, and it's ready to establish RP message communication again.

Use case - when Linux under uncontrolled reboot & what to do when Linux comes back up & sees status is not zero.

Action: Tanmay has to implement the 2 solutions on kernel mailing list & put links to those patches here.

Action: @glneo and/or Beleswar to review the Linux patch.  https://lore.kernel.org/all/20260317201251.3920841-1-tanmay.shah@amd.com/

Once Linux side is decided, we can have OpenAMP implemented accordingly.

nathalie-ckc:
2026-07-15 System Reference call:
Do we need to maintain backward compatibility?  Yes, if it exists upstream in Linux kernel -> It does exist in Xilinx kernel fork and Tanmay is trying to upstream to kernel.  So, better to wait on upstream Linux kernel to merge this patch.  
Tanmay will close this PR & open a new one once we have new mechanism in Linux kernel.

tnmysh:
Closing this PR, as it was discussed that no need to introduce current method of virtio device reset. It is in-complete, and linux kernel should update new method to reset the virtio device. After that new PR will be needed to detect the reset reqeust from the driver and reset the virtio device.

PR #101, opened 2026-03-20
examples: linux: add rpmsg udev rules and scripts

tnmysh:
Linux userspace apps can't use rpmsg devices without root access. Introduce udev rules to modify rpmsg device permissions so that userspace apps can use these devices. Also, create symlink to rpmsg device  with channel name, src ept and dest ept. Apps can use these symlinks directly without going through sysfs to find out which rpmsg device to use. This makes apps to use rpmsg devices very easy. It is also required to add current user to "rpmsg" group for apps to use rpmsg devices now. This also makes sure rpmsg communication is secure.

### DIFF
--- a/examples/linux/rpmsg_udev_scripts/99-rpmsg.rules
+++ b/examples/linux/rpmsg_udev_scripts/99-rpmsg.rules
@@ -1,0 +1,60 @@
+# Copyright (C) 2026, Advanced Micro Devices, Inc.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice,
+# this list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its contributors
+# may be used to endorse or promote products derived from this software without
+# specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+
+# Skip rpmsg_ctrl and rpmsg_ns devices by name using GOTO
+SUBSYSTEM=="rpmsg", ACTION=="add", \
+    KERNEL=="virtio*.rpmsg_ctrl.0.0", GOTO="rpmsg_end"
+
+SUBSYSTEM=="rpmsg", ACTION=="add", \
+    KERNEL=="virtio*.rpmsg_ns.53.53", GOTO="rpmsg_end"
+
+SUBSYSTEM=="rpmsg", ACTION=="add", \
+    KERNEL=="rpmsg_ctrl[0-9]*", GOTO="rpmsg_end"
+
+SUBSYSTEM=="rpmsg", ACTION=="remove", \
+    KERNEL=="rpmsg_ctrl[0-9]*", GOTO="rpmsg_end"
+
+# rpmsg endpoint creation rule
+SUBSYSTEM=="rpmsg", KERNEL=="rpmsg[0-9]*", ACTION=="add", \
+        GROUP="rpmsg", MODE="0660", \
+        RUN+="/usr/bin/rpmsg_add_ept_symlink.sh %k" GOTO="rpmsg_end"
+
+# rpmsg endpoint creation rule
+SUBSYSTEM=="rpmsg", KERNEL=="rpmsg[0-9]*", ACTION=="remove", \
+        RUN+="/usr/bin/rpmsg_remove_ept_symlink.sh %k" GOTO="rpmsg_end"
+
+# rpmsg channel creation rule
+SUBSYSTEM=="rpmsg", ACTION=="add", RUN+="/usr/bin/rpmsg_create_channel.sh %k"
+
+# ttyRPMSG device permission rule
+SUBSYSTEM=="tty", KERNEL=="ttyRPMSG[0-9]*", ACTION=="add", \
+    GROUP="rpmsg", MODE="0660"
+
+LABEL="rpmsg_end"
--- a/examples/linux/rpmsg_udev_scripts/README.md
+++ b/examples/linux/rpmsg_udev_scripts/README.md
@@ -1,0 +1,119 @@
+# RPMsg udev Scripts
+
+Udev rules and helper scripts to automatically manage RPMsg endpoint devices.
+When an RPMsg channel is created by the remote processor, these scripts handle
+device permissions, symlink creation, and endpoint export automatically.
+
+## Files
+
+| File | Description |
+|------|-------------|
+| `99-rpmsg.rules` | udev rules — triggers scripts on RPMsg char device add/remove events |
+| `rpmsg_create_channel.sh` | Creates RPMsg endpoint for a new channel using `rpmsg_export_ept` |
+| `rpmsg_add_ept_symlink.sh` | Creates user-friendly symlinks and sets device permissions on endpoint add |
+| `rpmsg_remove_ept_symlink.sh` | Removes symlinks when endpoint is removed |
+
+## How It Works
+
+When the remote processor starts and establishes an RPMsg channel, the kernel
+creates devices under `/sys/bus/rpmsg/devices/`. The udev rules trigger the
+scripts in the following order:
+
+```
+Remote processor boots
+    ↓
+RPMsg channel created → virtio0.rpmsg-openamp-demo-channel.-1.1024
+    ↓ udev ACTION==add
+rpmsg_create_channel.sh  → creates /dev/rpmsg0 via rpmsg_export_ept
+    ↓ udev ACTION==add (rpmsg0 endpoint)
+rpmsg_add_ept_symlink.sh → creates symlinks
+    ├── /dev/rpmsg-openamp-demo-channel.-1.1024  → /dev/rpmsg0
+    └── /dev/rpmsg0.rpmsg-openamp-demo-channel.-1.1024 → above symlink
+```
+
+On endpoint removal:
+```
+udev ACTION==remove (rpmsg0)
+    ↓
+rpmsg_remove_ept_symlink.sh → removes both symlinks
+
+
+```
+When removing the symlink, we need to find out which rpmsgX device is mapped
+to which symlink, and to track that the second symlink was created with name,
+rpmsgX.ch_name.ch_src.ch_dst. So, when the device is destroyed, we know which
+symlinks to remove.
+
+## Installation
+```bash
+# Copy scripts to /usr/bin
+sudo cp rpmsg_create_channel.sh   /usr/bin/
+sudo cp rpmsg_add_ept_symlink.sh  /usr/bin/
+sudo cp rpmsg_remove_ept_symlink.sh /usr/bin/
+sudo chmod +x /usr/bin/rpmsg_*.sh
+
+# Install udev rule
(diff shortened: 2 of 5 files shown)


examples/linux/rpmsg_udev_scripts/99-rpmsg.rules
```diff
@@ -0,0 +1,60 @@
...
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
```
arnopo: can we use BSD-3-Clause license  instead?

arnopo (commented):
I'm not  an udev expert, I need to play  with it but look nice in a first approach.

In addition of the licensing, I wonder if it should not be  a sub-folder of  `examples/linux/rpmsg-utils`. Indeed there is a dependency  with `rpmsg_utils`

what about putting it in `examples/linux/rpmsg-utils/udev` ?

examples/linux/rpmsg_udev_scripts/99-rpmsg.rules
tnmysh: Ack.

tnmysh:
> I'm not an udev expert, I need to play with it but look nice in a first approach.
> 
> In addition of the licensing, I wonder if it should not be a sub-folder of `examples/linux/rpmsg-utils`. Indeed there is a dependency with `rpmsg_utils`
> 
> what about putting it in `examples/linux/rpmsg-utils/udev` ?

I am still unsure if these scripts will work only for xlnx platform or it will be vendor agnostic ?
Same for udev rules. I plan to make it work in vendor agnostic manner for all rpmsg_char devices and  ttyRPMSG devices atleast, but need confirmation from other vendors.

If it works for all the vendors, then we can keep it under rpmsg-utils. If not, then probably create vendor directory and each vendor can have their own udev rules and directories.

edmooring (approved):
Once @arnopo comment is resolved. It looks good to me.

nathalie-ckc:
Discussion in 4/15/2026 OpenAMP System Reference call:
* Make the /dev names for the links more usable with prefix instead of having 2 symlinks

tnmysh:
@arnopo updated udev rules and scripts as per our previous discussion:

Changes in v2:
  - update rpmsg udev symlink handling
  - multiple retries to create rpmsg dev until name, src and dst files are
    available
  - make symlink format as:
    rpmsg_<ch name>.<src>.<dst>.rpmsgX where rpmsgX is /dev/rpmsgX.
  - Parse channel name from right-to-left, so endpoint numbers can be parsed
    correctly in case channel name contains dots too.
  - fix bsd-3-clause license text
  - add makefile to install udev rules script

tnmysh:
@arnopo if this works on multiple platforms, I don't mind moving it under `examples/linux/rpmsg-utils/udev` Please let me know your thoughts.

examples/linux/rpmsg_udev_scripts/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Requirements
+
+- Linux kernel 6.18 or later (`rpmsg_add_ept_symlink.sh`)
+- `rpmsg_export_ept` utility available in `/usr/bin`
+- udev running on the target system
+
+## Debugging
+
+# Monitor udev events
```
arnopo, line 123:
```suggestion
# Monitor udev events
+```bash
```

examples/linux/rpmsg_udev_scripts/rpmsg_create_channel.sh
```diff
@@ -0,0 +1,21 @@
...
+ch_src="${rpmsg_dev_rest##*.}"
+rpmsg_dev_rest="${rpmsg_dev_rest%.*}"
+virtio_name="${rpmsg_dev_rest%%.*}"
+ch_name="${rpmsg_dev_rest#*.}"
+
+#find control device for this channel e.g. /dev/rpmsg_ctrl0
+rpmsg_ctrl_dev=$(ls "/sys/bus/rpmsg/devices/$rpmsg_dev/subsystem/devices/$virtio_name.rpmsg_ctrl.0.0/rpmsg/")
+
+# create endpoint for this channel e.g. /dev/rpmsg0
+res=$(rpmsg_export_ept "/dev/$rpmsg_ctrl_dev" "$ch_name" "$ch_src" "$ch_dest")
```
arnopo, line 21: res seems not used
```suggestion
rpmsg_export_ept "/dev/$rpmsg_ctrl_dev" "$ch_name" "$ch_src" "$ch_dest"
```

examples/linux/rpmsg_udev_scripts/rpmsg_create_channel.sh
```diff
@@ -0,0 +1,21 @@
...
+# e.g. virtio0.rpmsg-openamp-demo-channel.-1.1024
+rpmsg_dev=$1
+
+# Split from the end so channel names can contain dots.
+ch_dest="${rpmsg_dev##*.}"
+rpmsg_dev_rest="${rpmsg_dev%.*}"
+ch_src="${rpmsg_dev_rest##*.}"
+rpmsg_dev_rest="${rpmsg_dev_rest%.*}"
+virtio_name="${rpmsg_dev_rest%%.*}"
+ch_name="${rpmsg_dev_rest#*.}"
```
arnopo, line 15: Additional check suggested by IA

```
if [ -z "$virtio_name" ] || [ -z "$ch_name" ] || [ -z "$ch_src" ] || [ -z "$ch_dest" ]; then
	echo "failed to parse rpmsg device name: $rpmsg_dev" >&2
	exit 1
fi

if ! command -v rpmsg_export_ept >/dev/null 2>&1; then
	echo "rpmsg_export_ept is not available in PATH" >&2
	exit 1
fi
```

examples/linux/rpmsg_udev_scripts/rpmsg_create_channel.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (C) 2026, Advanced Micro Devices, Inc.
+
+# e.g. virtio0.rpmsg-openamp-demo-channel.-1.1024
+rpmsg_dev=$1
+
```
arnopo, line 8: command usage helper seems missing in script. please find a proposal below that should also be adapted and added in other script 

proposal to put on top of the file:
```
usage() {
	echo "Usage: $0 <rpmsg-channel-device>"
	echo ""
	echo "Create an RPMsg endpoint for a channel device announced by the kernel."
	echo ""
	echo "Example:"
	echo "  $0 virtio0.rpmsg-openamp-demo-channel.-1.1024"
}

if [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
	usage
	exit 0
fi
````
then here: 
```suggestion
+if [ -z "$rpmsg_dev" ]; then
+	echo "missing rpmsg device name argument" >&2
+	exit 1
+fi
```

examples/linux/rpmsg_udev_scripts/README.md
```diff
@@ -0,0 +1,131 @@
...
+removes links matching `/dev/rpmsg_*.rpmsgX`.
+
+## Installation
+```bash
+# Install scripts and udev rule on a target system
+sudo make install
+
+# Reload udev rules
+sudo udevadm control --reload-rules
+sudo udevadm trigger
```
arnopo, line 52: I will put udev reload at the end of the installation chapter

examples/linux/rpmsg_udev_scripts/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+
+```
+When removing the symlink, we need to find out which rpmsgX device is mapped
+to which symlink. The symlink name ends with `.rpmsgX`, so the removal script
+removes links matching `/dev/rpmsg_*.rpmsgX`.
+
+## Installation
+```bash
+# Install scripts and udev rule on a target system
```
arnopo, line 47:
```suggestion
+ # Install scripts and udev rule on a local system:
+```bash
+ sudo make install
+ ```
+ # Install scripts and udev rule on a target system:
+```bash
+ make DESTDIR=<image-root> install
+ ```
```

examples/linux/rpmsg_udev_scripts/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+> **Note:** Users must be members of the `rpmsg` group to access `/dev/rpmsg*`
+> devices. Changes to group membership require logout/login or `newgrp rpmsg`
+> to take effect.
+
+## Symlink Naming Convention
+
+| Symlink | Example | Points to |
+|---------|---------|-----------|
+| `rpmsg_ept_name.src.dst.rpmsg_dev` | `/dev/rpmsg_rpmsg-openamp-demo-channel.-1.1024.rpmsg0` | `/dev/rpmsg0` |
```
arnopo, line 101: I don’t remember all the details of our OpenAMP meeting. Could you please explain the reasoning behind the naming?

It seems to me that the symlink name does not meet the need, and it may not be stable.

For example, on the remote side, you create two RPMsg raw devices with the channel names “dev1” and “dev2”. Depending on the order the remote-side RPMsg channels are created, the symlink could be:

* /dev/rpmsg_dev1.-1.1024.rpmsg0
* /dev/rpmsg_dev1.-1.1025.rpmsg1

I’m also not a fan of using a dot in the name. This does not seem to be a common practice under /dev.

What do you think would be a more stable naming approach: /dev/rpmsg_dev1 or simply /dev/dev1?

arnopo (changes requested):

examples/linux/rpmsg_udev_scripts/README.md
tnmysh: Ack.

examples/linux/rpmsg_udev_scripts/README.md
tnmysh: Ack.

examples/linux/rpmsg_udev_scripts/README.md
tnmysh: Following is the reasoning behind the naming convention:

1) We want to allow userspace apps to communicate to remote without needing the root access. The apps don't know which /dev/rpmsgX device maps to which rpmsg channels. But they can know the channel name, and ept information ahead of the time. So, the symlink name has  "**(ch_name).(src_ept).(dst_ept)"** in it.

2) **rpmsg_ prefix**: The udev scripts don't have to go through all the files under /dev/, so symlink has **rpmsg_** prefix. With that prefix, scripts and apps can easily know which files under /dev/ directory they should check.

3) **_rpmsgX suffix**: Two cores can create rpmsg channel with the same name, and ept information. In such case, we want to distinguish the second rpmsg channel symlink. That is why added _rpmsgX suffix to the symlink name because only one /dev/rpmsgX device exist per rpmsg channel.

If we don't want dots, then probably I can replace them with underscore ( _ ) or hypen ( - ). If I have to pick from both, I'd pick underscore. But I am open to any other option.

examples/linux/rpmsg_udev_scripts/README.md
tnmysh: Ack.

examples/linux/rpmsg_udev_scripts/rpmsg_create_channel.sh
tnmysh: Ack.

examples/linux/rpmsg_udev_scripts/rpmsg_create_channel.sh
tnmysh: Ack.

examples/linux/rpmsg_udev_scripts/rpmsg_create_channel.sh
tnmysh: Ack.

examples/linux/rpmsg_udev_scripts/README.md
arnopo: > Following is the reasoning behind the naming convention:
> 
> 1. We want to allow userspace apps to communicate to remote without needing the root access. The apps don't know which /dev/rpmsgX device maps to which rpmsg channels. But they can know the channel name, and ept information ahead of the time. So, the symlink name has  "**(ch_name).(src_ept).(dst_ept)"** in it.
> 2. **rpmsg_ prefix**: The udev scripts don't have to go through all the files under /dev/, so symlink has **rpmsg_** prefix. With that prefix, scripts and apps can easily know which files under /dev/ directory they should check.
> 3. **_rpmsgX suffix**: Two cores can create rpmsg channel with the same name, and ept information. In such case, we want to distinguish the second rpmsg channel symlink. That is why added _rpmsgX suffix to the symlink name because only one /dev/rpmsgX device exist per rpmsg channel.

For point 3, if two cores create the same RPMsg channel, could the addresses also be identical, since you would have two RPMsg instances?

> 
> If we don't want dots, then probably I can replace them with underscore ( _ ) or hypen ( - ). If I have to pick from both, I'd pick underscore. But I am open to any other option.

yes seems to me better.


Regarding the symlink, it is used by the application. The symlink name should be a unique name that stays the same after each boot, right?

If you use the EPT address for the symlink name, that depends on factors you cannot guarantee. The EPT addresses depend on the RPMsg channel binding order.

nathalie-ckc:
System reference call 8/26/26:
We have consensus on the rule, but the symlink part is still under discussion. Tanmay will create a PR for the udev rule so that can get merged first & we keep this one to have the history of the symlink discussion.

PR #107, opened 2026-05-12
legacy_apps: xlnx: fix build system for native cmake build

tnmysh:
OPENAMP_APP_NAME variable is needed for xilinx specific build tools. Native cmake build doesn't need this variable, and so if it's not found then do not fail, and include project machine direcotry. This allows native cmake build to continue.

### DIFF
--- a/examples/legacy_apps/machine/xlnx/CMakeLists.txt
+++ b/examples/legacy_apps/machine/xlnx/CMakeLists.txt
@@ -23,6 +23,8 @@ include(CheckSymbolExists)
 #set_property (GLOBAL PROPERTY MACHINE ${MACHINE})
 #set (PROJECT_MACHINE ${MACHINE})
 
+set (_vitis_build ON)
+
 get_property (OPENAMP_APP_NAME GLOBAL PROPERTY OPENAMP_APP_NAME)
 if(OPENAMP_APP_NAME STREQUAL "rpc_demo")
   set(_app rpc_demo)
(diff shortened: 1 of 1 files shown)


examples/legacy_apps/machine/xlnx/CMakeLists.txt
```diff
@@ -31,7 +31,8 @@ elseif(OPENAMP_APP_NAME STREQUAL "echo")
 elseif(OPENAMP_APP_NAME STREQUAL "matrix_multiply")
   set (_app matrix_multiplyd)
 else()
-  message(FATAL_ERROR "OPENAMP_APP_NAME not picked up")
+  add_subdirectory(${PROJECT_MACHINE})
```
arnopo, line 36: Already done here:
https://github.com/OpenAMP/openamp-system-reference/blob/main/examples/legacy_apps/machine/xlnx/CMakeLists.txt#L75

tnmysh: HI @arnopo that addition is only used when OPENAMP_APP_NAME variable is passed, which will be passed via Xilinx's tools (Vitis). 

This patch is needed if we want to build with native cmake flow, without Xilinx's tools.

Thanks,
Tanmay

arnopo: My mistake, I missed the return() just after that. That is one reason why I am not a fan of return() in CMake  :smile: 

could you add a comment in code and/or commit explaining why you skip the rest of the cmake?

An alternative to avoid the return (only a suggestion you can ignore):

```cmake
set(_build_demo ON)

if(OPENAMP_APP_NAME STREQUAL "rpc_demo")
  set(_app rpc_demo)
elseif(OPENAMP_APP_NAME STREQUAL "echo")
  set(_app rpmsg-echo)
elseif(OPENAMP_APP_NAME STREQUAL "matrix_multiply")
  set(_app matrix_multiplyd)
else()
  add_subdirectory(${PROJECT_MACHINE})
  set(_build_demo OFF)
endif()

if(_build_demo)
  message("OpenAMP: OPENAMP_APP_NAME: ${OPENAMP_APP_NAME}")

  [...]
endif()
```

tnmysh: Thanks. I liked the idea. I will just use different variable name.

i.e. %s/_build_demo/_vitis_build/

tnmysh: Updated the commit.

examples/legacy_apps/machine/xlnx/CMakeLists.txt
```diff
@@ -31,120 +33,125 @@ elseif(OPENAMP_APP_NAME STREQUAL "echo")
...
+  set (_vitis_build off)
```
arnopo:
```suggestion
  set (_vitis_build OFF)
```

arnopo (commented):
minor comment else LGTM

examples/legacy_apps/machine/xlnx/CMakeLists.txt
tnmysh: Done.

arnopo (approved):

arnopo:
@tnmysh could you rebase it on top of main branch to check if this solves the doc generation issue?

tnmysh:
> @tnmysh could you rebase it on top of main branch to check if this solves the doc generation issue?

@arnopo I think the read-docs error is still there.

edmooring (approved):
Looks good to go.

PR #110, opened 2026-07-07
Fix: avoid reuse of resource table between remote app activations

### LINKED ISSUE #109: [Bug] Remote app fails to create RPMSG endpoint from second activation onwards
Starting a remote app creates a rpmsg endpoint only at the first activation, from the second activation onwards the endpoint is not created.

Tested on Kria KR260 with Linux as host, running on A53 cores, and a baremetal `rpmsg-echo` on a core R5.

`platform_create_rpmsg_vdev(...)` call to `restore_initial_rsc_table(...)` @ _examples/legacy_app/machine/xlnx/zynqmp_r5/platform_info.c_ is the suspect.

mc-cat-tty:
Solves #109 

This fix has been tested on Kria KR260 with `rpmsg-echo` bare metal application.

### DIFF
--- a/examples/legacy_apps/machine/xlnx/zynqmp_r5/platform_info.c
+++ b/examples/legacy_apps/machine/xlnx/zynqmp_r5/platform_info.c
@@ -415,7 +415,7 @@ platform_create_rpmsg_vdev(void *platform, unsigned int vdev_index,
 	struct metal_io_region *shbuf_io;
 	int ret;
 
-	restore_initial_rsc_table();
+	// restore_initial_rsc_table();
 
 	rpmsg_vdev = metal_allocate_memory(sizeof(*rpmsg_vdev));
 	if (!rpmsg_vdev)


tnmysh:
The commit message needs more explanation. What is the problem, and how this patch solves it.

examples/legacy_apps/machine/xlnx/zynqmp_r5/platform_info.c
```diff
@@ -415,7 +415,7 @@ platform_create_rpmsg_vdev(void *platform, unsigned int vdev_index,
 	struct metal_io_region *shbuf_io;
 	int ret;
 
-	restore_initial_rsc_table();
+	// restore_initial_rsc_table();
```
arnopo, line 418: Please  remove the line instead of commenting it

arnopo (changes requested):
As mentioned by @tnmysh, please rework your commit message in following format

```
legacy_apps: xlnx:  avoid restore rsc table between remote app activations on zynqmp r5

[ explain the issue and the fix]

Signed-off-by:   Your name< Tanmay Shah <your email address>
```

arnopo:
Issue addressed in https://github.com/OpenAMP/openamp-system-reference/pull/111
I close this one

PR #115, opened 2026-08-03
Rpmsg multi services upstream cleanup

bentheredonethat:


examples/zephyr/rpmsg_multi_services/CMakeLists.txt
```diff
@@ -8,6 +8,10 @@ find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
 
 project(rpmsg_multi_services)
 
+if(CONFIG_OPENAMP_EXTERNAL_LINKER_SCRIPT)
```
tnmysh: Where is this variable used? Is it in AMD-Xilinx's downstream zephyr?

If so, this is not standard for other platforms. We need to discuss process to allow custom linker script use for every platform.

By default, each platform uses "SOC_LINKER_SCRIPT" variable. Can we override it via command line? and then in the default zephyr assign default linker script only if that variable is not set?

@arnopo do you have any opinion on this?

tnmysh: Or, Zephyr already provide options to use custom linker script we can use that:

CONFIG_HAVE_CUSTOM_LINKER_SCRIPT=y
CONFIG_CUSTOM_LINKER_SCRIPT="linker_xtensa_intel_adsp_cavs.ld"

We can pass CONFIG_HAVE_CUSTOM_LINKER_SCRIPT=y in the prj.conf, and CONFIG_CUSTOM_LINKER_SCRIPT="ld path" in the west command line.

What do you all think?

@bentheredonethat, @arnopo

examples/zephyr/rpmsg_multi_services/README.rst
```diff
@@ -39,6 +39,28 @@ Zephyr
...
+The application can consume a complete primary Zephyr linker script and
+device tree generated from an OpenAMP domain description. Pass both generated
+artifacts to the build:
+
+.. code-block:: console
+
+   west build -p always -b <target board> \
+     openamp-system-reference/examples/zephyr/rpmsg_multi_services -- \
+     -DDTS_SOURCE=/absolute/path/to/rpu_zephyr.dts \
+     -DCONFIG_OPENAMP_EXTERNAL_LINKER_SCRIPT='"/absolute/path/to/openamp-zephyr.ld"'
```
tnmysh: Here, we should pass CONFIG_CUSTOM_LINKER_SCRIPT directly. No need of another config option for the same purpose.

examples/linux/rpmsg-utils/rpmsg_ping.c
```diff
@@ -65,11 +65,12 @@ int main(int argc, char **argv)
 		fprintf(stderr, "failed to write endpoint  %s", argv[1]);
 		exit(1);
 	}
-	ret = read(fd, buffer, 256);
+	ret = read(fd, buffer, sizeof(buffer) - 1);
```
tnmysh: The size of buffer should come from rpmsg buffer size. Which I believe is fixed to 512 bytes fix for now. Why can't we keep it 512 bytes?

examples/linux/rpmsg-utils/rpmsg_ping.c
```diff
@@ -65,11 +65,12 @@ int main(int argc, char **argv)
...
 	if (ret < 0) {
 		fprintf(stderr, "failed to read endpoint %s", argv[1]);
 		exit(1);
 	}
+	buffer[ret] = '\0';
```
tnmysh: So, read on rpmsg device will always return 512. So, no point on NULL terminating it at the end of 512 bytes, I guess. Instead, we should always print only 512 bytes and stop printing if '\0' is found in between. 

Edit: So, above buffer[ret] = '\0' not needed.

examples/zephyr/rpmsg_multi_services/CMakeLists.txt
bentheredonethat: @tnmysh  i am fine to remove and build instead with

I am fine to remove this @tnmysh 

we can replace with

  west build -p always -b <board> \
    openamp-system-reference/examples/zephyr/rpmsg_multi_services -- \
    -DDTS_SOURCE=/absolute/path/to/rpu_zephyr.dts \
    -DCONFIG_HAVE_CUSTOM_LINKER_SCRIPT=y \
    -DCONFIG_CUSTOM_LINKER_SCRIPT='"/absolute/path/to/openamp-zephyr.ld"'

examples/zephyr/rpmsg_multi_services/README.rst
bentheredonethat: sure -- as noted above

examples/linux/rpmsg-utils/rpmsg_ping.c
```diff
@@ -65,11 +65,12 @@ int main(int argc, char **argv)
...
 	fprintf(stderr, "message received: \"%s\"\n", buffer);
```
tnmysh: If you really want, see if there something like fnprintf, which allwows to print fix number of bytes. If not, you can use snprintf.

examples/zephyr/rpmsg_multi_services/src/main_remote.c
```diff
@@ -36,6 +35,9 @@ LOG_MODULE_REGISTER(openamp_rsc_table, LOG_LEVEL_DBG);
 #define MAX_TTY_EPT  2
 #define MAX_RAW_EPT  2
 
+#define TTY_RESPONSE_PREFIX_LEN (sizeof("TTY 0: ") - 1U)
```
tnmysh: In response prefix we don't need device id i.e. "0".

tnmysh: Infact, I don't understand purpose of this. It doesn't match with the commit text.

examples/zephyr/rpmsg_multi_services/src/main_remote.c
```diff
@@ -378,9 +380,13 @@ void app_rpmsg_tty(void *arg1, void *arg2, void *arg3)
...
-				snprintf(tx_buff, 8, "TTY %d: ", i);
-				memcpy(&tx_buff[7], tty_msg[i].data, tty_msg[i].len);
-				rpmsg_send(&tty_ept[i], tx_buff, tty_msg[i].len + 8);
+				snprintf(tx_buff, TTY_RESPONSE_PREFIX_LEN + 1,
+					 "TTY %d: ", i);
+				memcpy(&tx_buff[TTY_RESPONSE_PREFIX_LEN],
+				       tty_msg[i].data, tty_msg[i].len);
+				rpmsg_send(&tty_ept[i], tx_buff,
+					   tty_msg[i].len +
+					   TTY_RESPONSE_PREFIX_LEN);
```
tnmysh: This whole hunk can be dropped. It's not related to the commit text, and not needed at all.

examples/zephyr/rpmsg_multi_services/src/main_remote.c
```diff
@@ -420,9 +427,24 @@ void app_rpmsg_raw(void *arg1, void *arg2, void *arg3)
...
+				}
+				snprintf(buff, RAW_RESPONSE_PREFIX_LEN + 1,
+					 "from ept 0x%04x: ", raw_ept[i].addr);
+				memcpy(&buff[RAW_RESPONSE_PREFIX_LEN],
+				       raw_msg[i].data, raw_msg[i].len);
+				buff[RAW_RESPONSE_PREFIX_LEN +
+				     raw_msg[i].len] = '\0';
+				rpmsg_sendto(&raw_ept[i], buff,
+					     raw_msg[i].len +
+					     RAW_RESPONSE_PREFIX_LEN + 1U,
```
tnmysh: Same for this hunk too. I don't understand why it's needed.

tnmysh: I guess, can you show what problem you are facing via logs that needs this commit?

examples/zephyr/rpmsg_multi_services/CMakeLists.txt
tnmysh: CONFIG_HAVE_CUSTOM_LINKER_SCRIPT=y have this in conf file,

and only pass the path from the command line.

examples/linux/rpmsg-utils/rpmsg_ping.c
bentheredonethat: Agreed on increasing the local receive buffer to 512 bytes. I’ll retain sizeof(buffer) in the read() call so the bound remains tied to the destination array rather than duplicating its size as a literal.

examples/linux/rpmsg-utils/rpmsg_ping.c
bentheredonethat: read() returns the actual RPMsg message length rather than always returning 512 bytes. However, I agree that modifying the buffer is unnecessary if the print is explicitly bounded.
                                
I’ll read up to sizeof(buffer) and use:
`fprintf(stderr, "message received: \"%.*s\"\n", (int)ret, buffer);`
This prints at most the number of bytes returned by read() and stops earlier if a NUL is present. I’ll remove buffer[ret] = '\0'.

examples/zephyr/rpmsg_multi_services/src/main_remote.c
bentheredonethat: "TTY 0: " was only being used to calculate the length of the existing "TTY %d: " prefix; it was not replacing the formatted endpoint ID. I agree that the macro makes this unnecessarily obscure, so I’ll remove it.

The underlying one-line correction is still needed: the payload is copied at offset 7, but the existing code sends len + 8, which includes one byte beyond the prefix and payload. I’ll reduce the hunk to the direct len + 8 to len + 7 correction and explicitly mention the TTY off-by-one in the commit message.

examples/zephyr/rpmsg_multi_services/src/main_remote.c
bentheredonethat: The failure is visible as trailing garbage:

```
message received: "from ept 0x0402: ping /dev/rpmsg0Eү??"
message received: "from ept 0x0001: ping /dev/rpmsg1E????"

```
The raw prefix is 17 bytes. The current memcpy() starts at offset 17 and overwrites the NUL written by snprintf(), but rpmsg_sendto() sends payload length + 18. It therefore includes one byte
beyond the prefix and payload. A maximum-sized incoming payload can also exceed the 512-byte response buffer once the prefix is added.

I’ll simplify the fix: use the return value from snprintf() as the actual prefix length, check prefix length + payload length against the response buffer, and transmit exactly that many bytes.
No NUL needs to be put on the wire because RPMsg is length-framed. The Linux utility will bound its output using the length returned by read().

examples/linux/rpmsg-utils/rpmsg_ping.c
bentheredonethat: there is not -- i have current latest as:
`fprintf(stderr, "message received: \"%.*s\"\n", (int)ret, buffer);`

bentheredonethat:
@tnmysh  i have updated per my findings and your review comments -- i have tested latest on hardware to confirm

examples/zephyr/rpmsg_multi_services/src/main_remote.c
```diff
@@ -233,7 +234,8 @@ int platform_init(void)
 	}
 
 	/* declare resource table region */
-	rsc_table_get(&rsc_table, &rcs_size);
+	rsc_table_get(&rsc_table_ptr, &rcs_size);
+	rsc_table = rsc_table_ptr;
```
tnmysh: I think this whole commit can be dropped too. rsc_table, and rsc_table_ptr both are (void *) same type. I don't know what is solved here.

if incompatible pointer type is not seen now, it shouldn't be seen with the original patch too.

Please check again, if this is needed or not. Could you paste original error?

bentheredonethat: ok can drop this commit - thanks @tnmysh

tnmysh (approved):

examples/zephyr/rpmsg_multi_services/src/main_remote.c
```diff
@@ -418,9 +417,23 @@ void app_rpmsg_raw(void *arg1, void *arg2, void *arg3)
...
+					rpmsg_release_rx_buffer(
+						&raw_ept[i], raw_msg[i].data);
+					raw_msg[i].len = 0;
+					raw_msg[i].data = NULL;
+					continue;
+				}
+				memcpy(&buff[ret],
+				       raw_msg[i].data, raw_msg[i].len);
+				rpmsg_sendto(&raw_ept[i], buff,
+					     (size_t)ret + raw_msg[i].len,
```
arnopo: suggestion suggestion to avoid to call `rpmsg_release_rx_buffer` twice times
```c
				if (ret < 0 || (size_t)ret >= sizeof(buff) ||
				    raw_msg[i].len > sizeof(buff) - (size_t)ret) {
					LOG_ERR("Raw message is too large: %zu",
						raw_msg[i].len);
				} else {
				 	memcpy(&buff[ret],
				    	   raw_msg[i].data, raw_msg[i].len);
					rpmsg_sendto(&raw_ept[i], buff,
						     (size_t)ret + raw_msg[i].len,
						     raw_msg[i].src);
				}

```

examples/linux/rpmsg-utils/rpmsg_ping.c
```diff
@@ -48,7 +48,7 @@ static void usage(void)
 
 int main(int argc, char **argv)
 {
-	char buffer[256];
+	char buffer[512];
```
arnopo: It works only if the RPMsg size is 512, but Ok as it the default size for now

arnopo (commented):
one minor comment, else LGTM

arnopo (approved):

PR #116, opened 2026-08-17
legacy_apps: xlnx: move cmake required version only if built with vitis

tnmysh:
The project has default cmake required version which is overridden with xlnx specific cmake list file. Use this new required version only if vitis tool is used, otherwise use the default required version.


bentheredonethat (approved):
LGTM

edmooring (commented):
Looks good to go.

arnopo (approved):

PR #117, opened 2026-08-18
examples: linux: add 'rpmsg_' prefix to userspace app names

tnmysh:
Current linux userspace app names are vague and do not show the true purpose of the application. Add 'rpmsg_' prefix to the application binary's name. This prefix clearly shows that the respective binary will be used to send/receive rpmsg protocol payload.


bentheredonethat (approved):
LGTM

nathalie-ckc:
8/26/26 system reference call:
Arnaud pointed out there may be impact on docs & open-amp CI, yet to confirm.
OK if we push this to next release - TBD.

arnopo (approved):
I did not find any impact on the documentation or CI, so it is okay for me.

